PR #370, opened 2012-09-27
Async make

daniel-walker:
Makes it possible for the user to run the checkers in the background. It will only be used if the user enabled it, and the syntax checker makes use of it.

### DIFF
--- a/syntax_checkers/html/tidy.vim
+++ b/syntax_checkers/html/tidy.vim
@@ -10,6 +10,9 @@
 "
 "============================================================================
 " TODO: join this with xhtml.vim for DRY's sake?
+
+let async_html_syntax_checker = 1
+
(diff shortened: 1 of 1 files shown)


lucascaton:
Hi there!

Does it make sense to re-open this issue as we now have [Vim 8](https://laravel-news.com/2016/09/vim-8-0-is-released/)'s async I/O support?
Perhaps something similar to what [neomake](https://github.com/neomake/neomake) does?

Cheers.

lcd047:
> Does it make sense to re-open this issue as we now have [Vim 8](https://laravel-news.com/2016/09/vim-8-0-is-released/)'s async I/O support?

No, but it would make sense for you to read the thread I mentioned above.

PR #811, opened 2013-10-08
Allow the path to a `.pylintrc` to be provided.

s0undt3ch:


ryneeverett:
I wasn't aware of args overriding either. Example for future searchers:

`let g:syntastic_python_pylint_args = '--rcfile=~/.config/pylint/pylintrc'`

PR #926, opened 2014-01-03
Add jshint as a checker for html (via a redirect)

hjwp:
Alternative PR using redirectey file.

There may be a better way of adding that `extract=auto` argument, rather than requiring it in the vimrc?

### DIFF
--- a/syntax_checkers/html/jshint-redirect.vim
+++ b/syntax_checkers/html/jshint-redirect.vim
@@ -1,0 +1,25 @@
+"============================================================================
+"File:        jshint-redirect.vim
+"Description: Syntax checking plugin that uses jshint
+"Maintainer:  LCD 47 <lcd047 at gmail dot com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"             
+" requires g:syntastic_html_jshint_args="--extract=always" to be set in vimrc
+"============================================================================
+
+if exists("g:loaded_syntastic_html_jshint_checker")
+    finish
+endif
+let g:loaded_syntastic_html_jshint_checker = 1
+
+runtime! syntax_checkers/javascript/jshint.vim
+
+call g:SyntasticRegistry.CreateAndRegisterChecker({
+    \ 'filetype': 'html',
+    \ 'name': 'jshint',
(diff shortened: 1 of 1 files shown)


iasj:
I've been reading the [Syntastic Manual][Syntastic], like you said to me                                                               
[here][myissue], and trying to get syntax check for HTML files. I have
jshint installed with [yaourt][yaourt], and it work well from command line. See:

[Syntastic]: https://github.com/scrooloose/syntastic/blob/master/doc/syntastic.txt
[myissue]:   https://github.com/scrooloose/syntastic/issues/1338
[yaourt]:    https://wiki.archlinux.org/index.php/yaourt

*index.html*
```html
<html>
    <body>
</html>
```
and, if i run from command line, 
i vim $ jshint index.html                                                                                                              

index.html: line 1, col 1, Expected an identifier and instead saw '<'.
index.html: line 1, col 1, Expected an assignment or function call and instead saw an expression.
index.html: line 1, col 2, Missing semicolon.
index.html: line 2, col 5, Expected an identifier and instead saw '<'.
index.html: line 2, col 6, Unrecoverable syntax error. (50% scanned).
 
but syntastic tells me nothing!

Here is my syntastic configuration:

```vim
set statusline+=%#warningmsg#                                                                                           
set statusline+=%{SyntasticStatuslineFlag()}
set statusline+=%*

let g:syntastic_always_populate_loc_list = 1
let g:syntastic_auto_loc_list = 1
let g:syntastic_check_on_open = 1
let g:syntastic_check_on_wq = 0

let g:syntastic_error_symbol = "✗"
let g:syntastic_warning_symbol = "⚠"
let g:syntastic_auto_jump = 1
let g:syntastic_loc_list_height = 5

let g:syntastic_html_checkers = ["jshint"]                                                                                             
```
and i also have
```vim
set tabstop=4 
```

in my vimrc

Yet. If i hit :SyntasticInfo inside index.html, then i get 
Syntastic version: 3.6.0-45 (Vim 704, Linux)
Info for filetype: html
Global mode: active
Filetype html is active
The current file will be checked automatically
Available checkers: jshint validator w3
Currently enabled checker: jshint

which means that jshint is enabled.

I'm trying to get a web development environment setted up. Syntastic
worked well with csslint (for CSS) and with jshint (for JavaScript).
I realy don't know what i'm doing wrong. Any help will be appreciated.

lcd047:
JSHint is a JavaScript checker.  When run against HTML files it checks the embedded JavaScript code rather than the HTML structure.  Running `jshint index.html` is essentially useless; syntastic is actually running `jshint --verbose --extract always index.html` instead.  Since you have no JavaScript in your `index.html`, there is no error.  That's all there is to it.

PR #1102, opened 2014-06-03
RFC: Add ghc-modi syntax plugin for haskell

wuzzeb:
ghc-modi is a version of ghc-mod which caches data between runs, improving the time to check for errors to under half a second, even for large projects.

This is still more of an RFC, because I am not sure if this should go as a new checker or if I should merge the code directly into ghc-mod plugin.  Also, is the change to SyntasticMake ok?  Lastly, I still need to add a command to kill the ghc-modi instance.

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -398,6 +398,7 @@ endfunction " }}}2
 "
 "a:options can contain the following keys:
 "    'makeprg'
+"    'makeoutput'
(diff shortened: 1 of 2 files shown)


sdx23:
@wuzzeb having this as an external checker would be really nice. Would you do this? Otherwise, if you don't mind I'll provide (and maintain) this, especially since it also fixes #1576

PR #1280, opened 2014-12-25
Add htmlhint for HTML

hail2u:
[HTMLHint][1]:
> HTMLHint is a Static Code Analysis Tool for HTML, you can use it with
> IDE or in build system.

This tool can install easily via npm, and works without internet
connections.

[1]: https://github.com/yaniswang/HTMLHint

### DIFF
--- a/syntax_checkers/html/htmlhint.vim
+++ b/syntax_checkers/html/htmlhint.vim
@@ -1,0 +1,40 @@
+"============================================================================
+"File:        htmlhint.vim
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Kyo Nagashima <hail2u@gmail.com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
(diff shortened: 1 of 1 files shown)


lcd047:
Are you sure you're using the HTMLHint that you linked to?  The HTMLHint that I installed with `npm` produces errors in a format that isn't matched by your `errorformat`, and it outputs colour codes even when `stdout` is not a terminal.  Basically, your checker doesn't work.

On a more general note, the HTMLHint you link to seems less than great, to put it politely.  For one, it can't be used on UNIX because of an EOL problem, and the issues reporting this (see [here](https://github.com/yaniswang/HTMLHint/issues/40) and [here](https://github.com/yaniswang/HTMLHint/issues/42)) have been open for month.  There are other issues as well (f.i. [this](https://github.com/yaniswang/HTMLHint/pull/33)), that seem to imply that the author either doesn't really understand what he's doing, or doesn't care enough about this to make sure it's actually useful to other people...

Sorry, but no.  Please maintain it as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external) if you insist on using it.

PR #1282, opened 2014-12-26
JSCS checker: change to the directory of the file being linted

bregenspan:
JSLint, JSCS, and ESLint all support detecting config files. JSLint and ESLint use an exact config if specified, otherwise they look for one in the folder of the file, traversing upwards if none is found.

But JSCS currently looks in the current working folder, meaning that if someone opens files in several projects with different JSCS settings, the correct JSCS config will only be honored if they manually change the current working directory first.

The JSCS maintainers are working on making the config detection closer to JSLint/ESLint's (https://github.com/jscs-dev/node-jscs/issues/559), but in the meantime, this patch can normalize the behavior of the JSCS checker to better match the other common JS checkers.

### DIFF
--- a/syntax_checkers/javascript/jscs.vim
+++ b/syntax_checkers/javascript/jscs.vim
@@ -28,6 +28,7 @@ function! SyntaxCheckers_javascript_jscs_GetLocList() dict
 
     return SyntasticMake({
         \ 'makeprg': makeprg,
+        \ 'cwd': expand('%:p:h'),
         \ 'errorformat': errorformat,
         \ 'subtype': 'Style',
         \ 'preprocess': 'checkstyle',


lcd047:
I don't really like this, for two reasons.

First, I'm not a fan of adding workarounds in syntastic for deficiencies in third party software that would be much easier addressed at the source.

Second, changing the working directory is not as harmless as it looks.  For one, you patch breaks when you're checking files with a relative path.  Fixing this would require either passing full paths to JSCS, or changing the relative paths to unqualified names.  Loclists involving full paths are ugly, as filenames take half of the screen width; and using unqualified names makes it tricky to handle errors in included files.

Either way, it isn't a nice solution.

bregenspan:
@lcd047 both of those points make sense, thank you for the reply! I'll wait for the real fix on the JSCS side.

PR #1290, opened 2015-01-08
Make tab-width and columnchar depend on expandtab

euoia:
This PR makes it so that when using the phpcs PHP linter, a couple of settings are dependent on whether expandtab is set. This fixes the use-case where tabs and not spaces are used to indent lines.

### DIFF
--- a/syntax_checkers/php/phpcs.vim
+++ b/syntax_checkers/php/phpcs.vim
@@ -15,12 +15,20 @@ if exists("g:loaded_syntastic_php_phpcs_checker")
 endif
 let g:loaded_syntastic_php_phpcs_checker = 1
 
+if !exists('g:syntastic_php_phpcs_tab_width')
+    let g:syntastic_php_phpcs_tab_width = 4
+endif
+
+if !exists('g:syntastic_php_phpcs_column_specifier')
+    let g:syntastic_php_phpcs_column_specifier = '%c'
+endif
+
 let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_php_phpcs_GetLocList() dict
-    if &expandtab
-        let args = '--tab-width=' . &tabstop
+    if g:syntastic_php_phpcs_tab_width
+        let args = '--tab-width=' . g:syntastic_php_phpcs_tab_width
     else
         let args = ''
     endif
@@ -31,7 +39,7 @@ function! SyntaxCheckers_php_phpcs_GetLocList() dict
 
     let errorformat =
         \ '%-GFile\,Line\,Column\,Type\,Message\,Source\,Severity%.%#,'.
-        \ '"%f"\,%l\,%c\,%t%*[a-zA-Z]\,"%m"\,%*[a-zA-Z0-9_.-]\,%*[0-9]%.%#'
+        \ '"%f"\,%l\,' . g:syntastic_php_phpcs_column_specifier . '\,%t%*[a-zA-Z]\,"%m"\,%*[a-zA-Z0-9_.-]\,%*[0-9]%.%#'
 
     return SyntasticMake({
         \ 'makeprg': makeprg,
--- a/syntax_checkers/php/phpcs.vim
+++ b/syntax_checkers/php/phpcs.vim
@@ -19,13 +19,21 @@ let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_php_phpcs_GetLocList() dict
+    if &expandtab
+        let args = '--tab-width=' . &tabstop
+        let columnchar = '%v'
+    else
+        let args = ''
+        let columnchar = '%c'
+    endif
+
     let makeprg = self.makeprgBuild({
-        \ 'args': '--tab-width=' . &tabstop,
-        \ 'args_after': '--report=csv' })
+        \ 'args_after': '--report=csv',
+        \ 'args': args })
 
     let errorformat =
         \ '%-GFile\,Line\,Column\,Type\,Message\,Source\,Severity%.%#,'.
-        \ '"%f"\,%l\,%v\,%t%*[a-zA-Z]\,"%m"\,%*[a-zA-Z0-9_.-]\,%*[0-9]%.%#'
+        \ '"%f"\,%l\,' . columnchar . '\,%t%*[a-zA-Z]\,"%m"\,%*[a-zA-Z0-9_.-]\,%*[0-9]%.%#'
 
     return SyntasticMake({
         \ 'makeprg': makeprg,
--- a/syntax_checkers/php/phpcs.vim
+++ b/syntax_checkers/php/phpcs.vim
@@ -21,10 +21,8 @@ set cpo&vim
 function! SyntaxCheckers_php_phpcs_GetLocList() dict
     if &expandtab
         let args = '--tab-width=' . &tabstop
-        let columnchar = '%v'
     else
         let args = ''
-        let columnchar = '%c'
     endif
 
     let makeprg = self.makeprgBuild({
@@ -33,7 +31,7 @@ function! SyntaxCheckers_php_phpcs_GetLocList() dict
 
     let errorformat =
         \ '%-GFile\,Line\,Column\,Type\,Message\,Source\,Severity%.%#,'.
-        \ '"%f"\,%l\,' . columnchar . '\,%t%*[a-zA-Z]\,"%m"\,%*[a-zA-Z0-9_.-]\,%*[0-9]%.%#'
+        \ '"%f"\,%l\,%c\,%t%*[a-zA-Z]\,"%m"\,%*[a-zA-Z0-9_.-]\,%*[0-9]%.%#'
 
     return SyntasticMake({
         \ 'makeprg': makeprg,


lcd047:
Sorry, but this is wrong.  `%c` and `%v` in `errorformat` are supposed to be used when the external checker counts columns as bytes and "virtual" columns (that is columns after expanding tabs), respectively.  This has to do only with the external checker, and when done properly it works regardless of the setting of `expandtab`.  Now, the problem with `phpcs` is that it does or does not expand tabs according to the set of rules in use, and Vim has no way of telling when it's supposed to use `%c`, and when `%v`.  Syntastic uses `%v` simply because rule sets that count virtual columns seem to be more popular than rule sets that count bytes, but there's no way to make this work always.  Using `expandtab` as a toggle between `%c` and `%v` is not a solution that works for everybody either, it just changes the set of people that see it break.  If you want to see this fixed for good I'd suggest you convince `phpcs` developers to make the output of `phpcs` consistent. *shrug*

euoia:
Hi @lcd047 - thanks for looking at this.

After further testing it seems to me that simply using `%c` is sufficient. I can't find any case where `%v` is required.

Do you disagree with the change to make the `--tab-width` argument based on `expandtab`? If so perhaps we can just make this a user-configurable option with a default of the current behaviour.

What do you think?

lcd047:
> After further testing it seems to me that simply using `%c` is sufficient. I can't find any case where `%v` is required.

Example: `CakePHP` rules needs `%v`.

> Do you disagree with the change to make the `--tab-width` argument based on `expandtab`?

I do.

euoia:
In that case, do you have a better suggestion than making both `--tab-width` and `columchar` user-configurable?

lcd047:
For `--tab-width`, my suggestion is to leave it alone.  Removing it doesn't accomplish anything useful for `%c`, and breaks parsing for `%v`.

For the tabs vs. spaces problem, I suggest you complain to the `phpcs` developers.  I'm actually serious, solving this brain damage at the source would be much better than adding yet another non-standard option to syntastic.  For reference, dealing with this kind of stuff takes me 90% of time whenever I touch the checker code...

euoia:
I'll try to approach the problem in another way, but following your logic of minimising the number of non-standard options I think it would be better to remove the --tab-width=4 argument entirely from this checker.

lcd047:
By "standard options" I man the options generated by `makeprgBuild()`, such as `g:syntastic_*_args` and `g:syntastic_*_post_args`.  By "non-standard options" I mean ah-hoc options used by a single checker, such as `g:syntastic_perl_lib_path`.

As for removing `--tab-width`, please read again what I wrote above.

euoia:
I understood what you mean by standard options. My point is that by specifying an option like `--tab-width` on the command line, you're diverging from the default mode of operation of the tool. Even if I manage to get `phpcs` to allow `tab-width` to be specified in the ruleset, I'd expect the command line option will override it.

The specific problem I have with syntastic specifying the tab-width is that my files are tab-indented and use the PSR2.Methods.FunctionCallSignature rule with an indent value of "1" (as-in, 1 tab). When my files are converted to spaces, phpcs expects an additional indent of 1 space (which they don't have).

What's the reason for specifying `--tab-width=4` in the first place?

lcd047:
> What's the reason for specifying `--tab-width=4` in the first place?

It's `--tab-width=&tabstop` rather than `--tab-width=4`.  As I said above, parsing errors with `%v` is useless if Vim and phpcs disagree about the value of tabstop.  Since Vim can't find out phpcs's idea of tabstop but can force it to it's own `&tabstop`, it passes `--tab-width=&tabstop`.  The underlying assumption is that phpcs uses the value of tabstop only for computing column numbers in the reporter, rather than in the rules themselves.  Your use case breaks this last assumption. *shrug*

Now, if phpcs would just have an option to report column numbers as byte offsets rather than try to expand tabs, there would be no ambiguity, no need to agree on tabstop values, and syntastic could happily use `%c` instead of `%v` to parse errors.  The problem is, as I said above (several times already), that phpcs does or does not expand tabs according to the rules in use, and syntastic has no way to tell when it does and when it doesn't.  This can't be fixed in syntastic, it has to be fixed in phpcs.

euoia:
Thanks for spending so long on this. I will try to add an option to phpcs to report errors as byte offsets.

I just have one question. You've said that phpcs may or may not expand tabs based on the rules in use. I  haven't found any case where phpcs expands tabs except when `--tab-width` is used. I am using phpcs 2.0.0 and I tried with cakephp rulesets `2.0.x-dev` and `1.0.0` in composer. Can you please let me know how to reproduce that behaviour?

It looks like there was changes with respect to --tab-width behaviour from 1.x to 2.x - perhaps this is related? https://github.com/squizlabs/PHP_CodeSniffer/issues/340

lcd047:
> Can you please let me know how to reproduce that behaviour?

Take this file named `test.php`:
```php
<?php
class Foo {
        private $__foo = 77;
}
```
There's a tab at the beginning of the third line, and another tab between `private` and `$__foo`.
```
$ phpcs --standard=CakePHP test.php --report=csv File,Line,Column,Type,Message,Source,Severity
"/home/lcd047/test.php",3,1,error,"Tabs must be used to indent lines; spaces are not allowed",Generic.WhiteSpace.DisallowSpaceIndent.TabsUsed,5
"/home/lcd047/test.php",3,17,error,"Expected 1 blank line before member var; 0 found",Squiz.WhiteSpace.MemberVarSpacing.Incorrect,5

```
Please note column 17 for the second error.  With byte offsets that would have been 10.

This is with:
```
$ phpcs --version
PHP_CodeSniffer version 1.5.3 (stable) by Squiz (http://www.squiz.net)
```

I haven't tried with phpcs 2.x; actually I haven't paid much attention to anything PHP in a while.

> It looks like there was changes with respect to --tab-width behaviour from 1.x to 2.x - perhaps this is related? https://github.com/squizlabs/PHP_CodeSniffer/issues/340

Yes, I suppose that's it.  Sadly, the owner [says](https://github.com/squizlabs/PHP_CodeSniffer/issues/340#issuecomment-65869243):

> I can't change this back. It is a very important part of the way version 2.0 works.

lcd047:
I upgraded phpcs to 2.1.0.  Apparently CakePHP_CodeSniffer requires phpcs 1.4 or 1.5:
```
$ pear upgrade cakephp/CakePHP_CodeSniffer
cakephp/CakePHP_CodeSniffer requires package "pear/PHP_CodeSniffer" (version >= 1.4, version <= 1.5), installed version is 2.1.0
No valid packages found
upgrade failed
```

euoia:
Which version of CakePHP_CodeSniffer is that? The latest version in PEAR is 0.1.31 http://pear.cakephp.org/. I tried with that version (installed using composer) and it worked with phpcs 2.0.0 but I didn't get the same output as you.

By contrast composer has versions 0.1.31, 1.0.0 and 2.0.x-dev:
```
$ ./composer.phar show cakephp/cakephp-codesniffer
name     : cakephp/cakephp-codesniffer
descrip. : CakePHP CodeSniffer Standards
keywords : codesniffer, framework
versions : dev-master, 2.0.x-dev, 1.x-dev, 1.0.0, * 0.1.31, 0.1.30, 0.1.29, 0.1.28, 0.1.27, 0.1.26, 0.1.25, 0.1.24, dev-phpcs-fixer, dev-undercore-prefix, dev-ignore-underscore-prefix, dev-update-fxer-2, dev-update-fixer
type     : library
license  : MIT
source   : [git] https://github.com/cakephp/cakephp-codesniffer.git 520b4125468fcd6bb2cb6aaae3858fe77b30bf3b
dist     : [zip] https://api.github.com/repos/cakephp/cakephp-codesniffer/zipball/520b4125468fcd6bb2cb6aaae3858fe77b30bf3b 520b4125468fcd6bb2cb6aaae3858fe77b30bf3b
names    : cakephp/cakephp-codesniffer

requires
squizlabs/php_codesniffer ~1.4

requires (dev)
phpunit/phpunit 4.1.*
```

lcd047:
```
$ pear list -c cakephp
Installed packages, channel pear.cakephp.org:
=============================================
Package             Version State
CakePHP             2.6.0   stable
CakePHP_CodeSniffer 0.1.23  stable
```

euoia:
Ah that version is not even available in composer.

How do you want to proceed? I don't think the present situation is satisfactory. My preference would be to remove the tab-width argument entirely, but clearly this isn't a BC change.

One option would be a separate syntax checker for phpcs2. That would avoid the need for a non-standard option. I don't know which approach minimises the work on your end.

lcd047:
> Ah that version is not even available in composer.

Yay, lucky me.

> How do you want to proceed?

Well I can make syntastic require phpcs 2.0 or later, but that doesn't solve the problem:
```
$ phpcs --standard=CakePHP test.php --report=csv
File,Line,Column,Type,Message,Source,Severity,Fixable
"/home/lcd047/test.php",3,17,error,"Expected 1 blank line before member var; 0 found",Squiz.WhiteSpace.MemberVarSpacing.Incorrect,5,1
```
(the point being, it's still possible to fuckup column numbers with a "suitable" choice of rules).

euoia:
What versions of phpcs and CakePHP_CodeSniffer are you using to get that output? With 2.0.0 and 0.1.31 respectively, the column is reported as 11 when I run the same thing:

```
$ ./bin/phpcs --standard=CakePHP test.php --report=csv
File,Line,Column,Type,Message,Source,Severity
"/home/euoia/Code/copper/test.php",3,11,error,"Expected 1 blank line before member var; 0 found",Squiz.WhiteSpace.MemberVarSpacing.Incorrect,5
$ ./bin/phpcs --version
PHP_CodeSniffer version 1.5.6 (stable) by Squiz (http://www.squiz.net)
$ grep cakephp composer.json
        "cakephp/cakephp-codesniffer": "0.1.31"
```

lcd047:
> What versions of phpcs and CakePHP_CodeSniffer are you using to get that output?

The supposedly broken 0.1.23.  It doesn't matter, I still get tab expansions without CakePHP:
```
$ phpcs test.php --report=csv
File,Line,Column,Type,Message,Source,Severity,Fixable
"/home/lcd047/test.php",1,1,error,"Class found in \".php\" file; use \".inc\" extension instead",Squiz.Files.FileExtension.ClassFound,5,0
"/home/lcd047/test.php",1,1,error,"Missing file doc comment",Squiz.Commenting.FileComment.Missing,5,0
"/home/lcd047/test.php",2,1,error,"Missing class doc comment",Squiz.Commenting.ClassComment.Missing,5,0
"/home/lcd047/test.php",2,1,error,"Class name doesn't match filename; expected \"class test\"",Squiz.Classes.ClassFileName.NoMatch,5,0
"/home/lcd047/test.php",2,7,error,"Expected 1 space after class name; 7 found",Squiz.Classes.ClassDeclaration.SpaceAfterName,5,1
"/home/lcd047/test.php",2,10,error,"Spaces must be used for alignment; tabs are not allowed",Generic.WhiteSpace.DisallowTabIndent.NonIndentTabsUsed,5,1
"/home/lcd047/test.php",2,17,error,"Opening brace of a class must be on the line after the definition",Squiz.Classes.ClassDeclaration.OpenBraceNewLine,5,1
"/home/lcd047/test.php",3,1,error,"Spaces must be used for alignment; tabs are not allowed",Generic.WhiteSpace.DisallowTabIndent.NonIndentTabsUsed,5,1
"/home/lcd047/test.php",3,16,error,"Spaces must be used for alignment; tabs are not allowed",Generic.WhiteSpace.DisallowTabIndent.NonIndentTabsUsed,5,1
"/home/lcd047/test.php",3,17,error,"Member variable \"__foo\" is not in valid camel caps format",Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps,5,0
"/home/lcd047/test.php",3,17,error,"Missing member variable doc comment",Squiz.Commenting.VariableComment.Missing,5,0
"/home/lcd047/test.php",3,17,error,"Expected 1 blank line before member var; 0 found",Squiz.WhiteSpace.MemberVarSpacing.Incorrect,5,1
"/home/lcd047/test.php",4,1,error,"Expected //end class",Squiz.Commenting.ClosingDeclarationComment.Missing,5,1
```
This is with phpcs 2.1.0. *shrug*

euoia:
That's so odd. With phpcs 2.1.0 installed from composer I get different output:
```
$ ls -1
composer.json
composer.phar
test.php
$ cat composer.json
{
    "require": {
        "squizlabs/php_codesniffer": "2.1.0"
    },
    "config": {
        "bin-dir": "bin/"
    }
}
$ ./composer.phar update
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing squizlabs/php_codesniffer (2.1.0)
    Loading from cache

Writing lock file
Generating autoload files
$ ./composer.phar install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Generating autoload files
$ phpcs test.php --report=csv
File,Line,Column,Type,Message,Source,Severity,Fixable
"/home/euoia/Code/test-phpcs/test.php",2,1,error,"Missing file doc comment",PEAR.Commenting.FileComment.Missing,5,0
"/home/euoia/Code/test-phpcs/test.php",2,1,error,"Missing class doc comment",PEAR.Commenting.ClassComment.Missing,5,0
"/home/euoia/Code/test-phpcs/test.php",2,11,error,"Opening brace of a class must be on the line after the definition",PEAR.Classes.ClassDeclaration.OpenBraceNewLine,5,1
"/home/euoia/Code/test-phpcs/test.php",3,1,error,"Spaces must be used to indent lines; tabs are not allowed",Generic.WhiteSpace.DisallowTabIndent.TabsUsed,5,1
"/home/euoia/Code/test-phpcs/test.php",3,2,error,"Line indented incorrectly; expected at least 4 spaces, found 1",PEAR.WhiteSpace.ScopeIndent.Incorrect,5,1
"/home/euoia/Code/test-phpcs/test.php",3,9,error,"Spaces must be used for alignment; tabs are not allowed",Generic.WhiteSpace.DisallowTabIndent.NonIndentTabsUsed,5,1
$ cat test.php
<?php
class Foo {
        private         $__foo = 77;
}
```

Be aware that phpcs also have a configuration file that is uses to store values set with `--config-set`. The configuration file is typically stores in `./vendor/squizlabs/php_codesniffer/CodeSniffer.conf`. You can see your configuration values with `./bin/phpcs --config-show`.

euoia:
If you run with -v what standard is being used?
```
$ ./bin/phpcs -v test.php --report=csv
Registering sniffs in the PEAR standard... DONE (27 sniffs registered)
Creating file list... DONE (1 files in queue)
Changing into directory /Users/james.pickard/Code/copper
Processing test.php [PHP => 19 tokens in 4 lines]... DONE in 4ms (6 errors, 0 warnings)
...
```

lcd047:
```
$ phpcs -v test.php --report=csv
Registering sniffs in the MySource standard... DONE (143 sniffs registered)
Creating file list... DONE (1 files in queue)
Changing into directory /home/lcd047
Processing test.php [PHP => 19 tokens in 4 lines]... DONE in 1ms (11 errors, 0 warnings)
File,Line,Column,Type,Message,Source,Severity,Fixable
[...]

$ strace -eopen phpcs -v test.php --report=csv 2>&1 | fgrep '.conf' | fgrep -vw ENOENT
open("/usr/lib/php/data/PHP_CodeSniffer/CodeSniffer.conf", O_RDONLY) = 3

$ cat /usr/lib/php/data/PHP_CodeSniffer/CodeSniffer.conf
<?php
 $phpCodeSnifferConfig = array (
  'default_standard' => 'MySource',
  'tab_width' => '8',
)
?>
```

lcd047:
Actually, I can't reproduce your results even if I choose the PEAR rules:
```
$  phpcs -v --standard=PEAR test.php --report=csv
Registering sniffs in the PEAR standard... DONE (27 sniffs registered)
Creating file list... DONE (1 files in queue)
Changing into directory /home/lcd047
Processing test.php [PHP => 19 tokens in 4 lines]... DONE in 1ms (5 errors, 0 warnings)
File,Line,Column,Type,Message,Source,Severity,Fixable
"/home/lcd047/test.php",2,1,error,"Missing file doc comment",PEAR.Commenting.FileComment.Missing,5,0
"/home/lcd047/test.php",2,1,error,"Missing class doc comment",PEAR.Commenting.ClassComment.Missing,5,0
"/home/lcd047/test.php",2,11,error,"Opening brace of a class must be on the line after the definition",PEAR.Classes.ClassDeclaration.OpenBraceNewLine,5,1
"/home/lcd047/test.php",3,1,error,"Spaces must be used to indent lines; tabs are not allowed",Generic.WhiteSpace.DisallowTabIndent.TabsUsed,5,1
"/home/lcd047/test.php",3,16,error,"Spaces must be used for alignment; tabs are not allowed",Generic.WhiteSpace.DisallowTabIndent.NonIndentTabsUsed,5,1
```

euoia:
The reason you're getting different column positions is because you are using a config file which sets a tab-width of 8.

I don't know where /usr/lib/php/data/PHP_CodeSniffer/CodeSniffer.conf comes from but it doesn't exist on my system. There are few references to it on the web. Is it possible you've use ran phpcs --set-config to set the tab-width to 8 in the past?

According to this page, the default mode of operation should not convert spaces to tabs:
http://pear.php.net/manual/en/package.php.php-codesniffer.config-options.php

lcd047:
You're right, deleting `tab-width` from that file fixed the problem, which is exactly the behaviour mentioned in the manual.  Ok, I removed `--tab-width=&tabstop` and changed `%v` to `%c` in syntastic: be01cac.

I still think the output of phpcs should be made consistent, either by always reporting column numbers as byte offsets, or by adding separate fields for columns and virtual columns to the output.  I suppose always passing `--tab-width` was safer, in that it made phpcs always report virtual columns.  Sadly, this breaks with phpcs 2.x since `tab-width` is now _also_ used for rules.  And syntastic still has no way to tell for sure whether it should use `%c` or `%v`.  Maybe the next time somebody reports a problem about that and asks me to replace `%c` with `%v` the right approach is to remove the support for parsing columns numbers altogether...

euoia:
Thanks for making this change and for your responses throughout the thread. Syntastic is a very useful plugin. I'll have a look through the issues and see if there's any way I can contribute.

PR #1291, opened 2015-01-09
Make go.vim work with go 1.4 again

riannucci:
The addition of 'opts' caused this code to render bad command lines like `go build '' -o /dev/null`, which caused go 1.4 to treat `-o` as a package, and then fail (producing no error output for syntastic :(). I didn't test this with 1.3, but I expect that it should still work.


riannucci:
BTW, I suspect the breakage happened here: scrooloose/syntastic@a18ccb40df

lcd047:
No, adding an empty argument is not what's supposed to happen there.  Please try 74a0713 instead.

PR #1294, opened 2015-01-14
Rubocop improvements

p0deje:
This pull request brings a couple of things here:

1. Removes `--silent` option since Rubocop >= 0.12 dropped it (and in [master](https://github.com/bbatsov/rubocop/blob/master/CHANGELOG.md#changes) throws error.
2. Allows to pass custom options to Rubocop (in my case it's `--display-cop-names` which is useful when initial config is being created)

If this is merged, I'll update wiki page accordingly.


lcd047:
Thank you.  I included the first part, dropping `--silent`, in 8b97caa.

The second part, creating a custom variable to pass arguments to rubocop is not needed, setting `g:syntastic_ruby_rubocop_args` should already work.

p0deje:
Thanks, I think I should have read doc better!

PR #1298, opened 2015-01-17
Use codecs to read the python file

delfick:
Hello,

I use this testing library I created called noseOfYeti (https://noseofyeti.readthedocs.org/en/latest/)

This library means that my tests use a codec that transforms the file at import time.

This pull request is to make the python checker aware of the codec (works generically with any codec specified at the top of a file) and syntax check the decoded file, rather than the raw file.

Stephen.

### DIFF
--- a/syntax_checkers/python/compile.py
+++ b/syntax_checkers/python/compile.py
@@ -2,12 +2,29 @@
 
 from __future__ import print_function
 from sys import argv, exit
-
+import codecs
+import re
+import os
 
 if len(argv) != 2:
     exit(1)
 
 try:
-    compile(open(argv[1]).read(), argv[1], 'exec', 0, 1)
+    with open(argv[1]) as fle:
+        text = fle.readlines()
+
+    if text:
+        match = re.match("#\s*coding\s*:\s*(?P<coding>\w+)", text[0])
+        if match:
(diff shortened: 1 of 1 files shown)


lcd047:
> syntax check the decoded file, rather than the raw file

Not sure why so many people try to do this (that is, preprocess the file in some way and check the transformed file), and expect the results to still be meaningful to syntastic.  The line and column numbers you get from checking the decoded file refer to (surprise!) the decoded file; syntastic needs line and column coordinates for the __original__ file.

delfick:
One of the features of noseOfYeti in particular is that the line numbers of the result matches the original file.

lcd047:
But not the column numbers on the lines it replaces.  Well, this is a pretty specialised use case anyway, I suppose it would be fine as a separate checker.

lcd047:
... Or perhaps an option to the existing python checker to use an alternate `compile.py`.

delfick:
That's a good option

lcd047:
Ok, done in 5359f7f.  Please update the corresponding [wiki page](https://github.com/scrooloose/syntastic/wiki/Python%3A---python).

delfick:
Awesome. Thanks for that!

I've updated the wiki page.

PR #1311, opened 2015-01-30
add swift syntax checker support using xcrun

tomfogg:


lcd047:
Please add a licence to your checker.

tomfogg:
added licence

lcd047:
Thank you.  Merged, with this patch on top: b9608e7.

PR #1313, opened 2015-01-31
missing comma in swift errorformat after cleanup

tomfogg:


lcd047:
Right, sorry about that.

PR #1317, opened 2015-02-09
Expose xmllint arguments through a global variable

ametheus:
At my organisation we don't use DTD's, which leads to hundreds of validity errors (e.g. "no declaration for element div") per file. (In fact, about one per line.)

I've added the ability to override the CLI arguments for xmllint, so my vimrc reads:
```vim
let g:syntastic_xml_xmllint_cli_args = '--noout'

" [...]
execute pathogen#infect('bundle/{}')
```


lcd047:
Ok, done (properly) in 3ea613f.  Please read the manual and see the [wiki](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#makeprgbuildoptions) for details.

ametheus:
Sweet, thanks!

PR #1323, opened 2015-02-13
Don't warn on python/prospector with no messages.

Shizmob:
When the python/prospector checker finds no issues in the file, the "messages" key is absent in the JSON output. Make sure we don't spew out a warning when that happens.


lcd047:
Right, thank you.

PR #1329, opened 2015-02-18
Check maven file path with slashes

trygveaa:
When finding the output directory for maven, check the path for the
current file with parts separated by slashes, not just dots
(src/main/java in addition to src.main.java).

It is common in maven do have one subdirectory per part, so we should
support this.

### DIFF
--- a/syntax_checkers/java/javac.vim
+++ b/syntax_checkers/java/javac.vim
@@ -388,13 +388,15 @@ function! s:MavenOutputDirectory() " {{{2
         if has_key(mvn_properties, 'project.properties.build.dir')
             let output_dir = mvn_properties['project.properties.build.dir']
         endif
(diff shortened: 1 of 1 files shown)


lcd047:
@troydm Any comments?

troydm:
source of this statement?
According to [this](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) there are no references to dotted path

lcd047:
I see; fixed in cf4e18d.

While you are at it: are nearby `let output_dir .= '/target/classes'` and `let output_dir .= '/target/test-classes'` ok, or `/` should be OS-sensitive?

troydm:
@lcd047 Yeah, it should be OS-sensitive, can you fix that too while you are at it?

lcd047:
Ok, done in d73546d.

PR #1339, opened 2015-02-25
Add jsxcs checker

mgechev:
When using [jscs](https://www.npmjs.com/package/jscs) with ReactJS application, which declares the markup of the component through JSX, there might be some invalid errors.

This is a syntax checker for [jsxcs](https://www.npmjs.com/package/jsxcs), which solves the issue.


lcd047:
... Or just add this to your vimrc:
```vim
let g:syntastic_javascript_jscs_exec = 'jsxcs'
```

mgechev:
Same for `jshint` and `jsxhint`.

lcd047:
Nope, the corresponding checkers are pretty different.

PR #1341, opened 2015-02-26
Using flow server instead of `flow check`

verkholantsev:
`flow check` is too slow on large projects, so it is preferable to use flow server with `flow status` command.


lcd047:
Does this need a different setup than `flow check`?  If it does, it has to be documented in the wiki.

lcd047:
Actually, I'd prefer it to be configurable.  Commit 413cdb3.

PR #1342, opened 2015-03-02
Interaction with vim-auto-save is fixed now

907th:
... in this commit:
https://github.com/907th/vim-auto-save/commit/24b491de94f7ea799663f97f7f4002e6ae87b0fd


lcd047:
Thank you.  Manual updated in 272fc7d.

PR #1344, opened 2015-03-06
flow js checker - use 'flow' instead of 'flow check' for speed

seanhess:
I changed the command from "flow check" to "flow"

Right now, my editor freezes for 2-3 seconds every time I save, because that's how long it takes to boot up the flow type checker. Flow is designed to turn on a type checking server once, which keeps track of file changes on its own. You run `flow` and if the server isn't on, it turns it on. Then you just run `flow` again and it reports updates. 

This PR removes "check", which makes subsequent type checks instant. The first time you save, it still freezes, but from then on it's great. I verified that it creates the server in the normal way (running flow in the terminal after saving doesn't start a second server).

### DIFF
--- a/syntax_checkers/javascript/flow.vim
+++ b/syntax_checkers/javascript/flow.vim
@@ -23,7 +23,7 @@ set cpo&vim
 
 function! SyntaxCheckers_javascript_flow_GetLocList() dict
     let makeprg = self.makeprgBuild({
-        \ 'exe': self.getExecEscaped() . ' check',
+        \ 'exe': self.getExecEscaped(),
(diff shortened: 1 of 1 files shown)


gabelevi:
Cool! If you don't want to make the user wait for the server to start, you can also do something like `--timeout 2` and the server will exit with error code `3` if it takes more than 2 seconds.

lcd047:
That's great, except (1) somebody else suggested something similar a few days ago (see #1341), and (2) the patch is not needed:
```vim
let g:syntastic_javascript_flow_exe = 'flow'
```
(This is not to be confused with `g:syntastic_javascript_flow_exec`).

PR #1349, opened 2015-03-10
Fixed 'NUL' file being created on Windows.

eyeofhell:
Unfortunately, on Windows TSC doesn't support 'NUL' as '--out' destination. Fixed this by detecting windows.


lcd047:
Please see #1155.  The problem is in node (or rather in libuv), please get it fixed there.  I may consider a workaround in syntastic if you provide evidence node developers refuse to fix the problem, but not until then.  Sorry, but I'm not a fan of cleaning up other people's mess. _shrug_

PR #1362, opened 2015-03-17
add IsAvailable to rst2pseudoxml syntax check

mmv:
If IsAvailable is absent, this checker will break when the program isn't available

Note: this needs further testing (for the case when the program *is* available).

### DIFF
--- a/syntax_checkers/rst/rst2pseudoxml.vim
+++ b/syntax_checkers/rst/rst2pseudoxml.vim
@@ -23,6 +23,14 @@ let s:rst2pseudoxml = executable('rst2pseudoxml.py') ? 'rst2pseudoxml.py' : 'rst
 let s:save_cpo = &cpo
 set cpo&vim
 
+function! SyntaxCheckers_rst_rst2pseudoxml_IsAvailable() dict
+    if !executable(s:rst2pseudoxml)
+        return 1
+    endif
+
+    return 0
+endfunction
+
 function! SyntaxCheckers_rst_rst2pseudoxml_GetLocList() dict
     let makeprg = self.makeprgBuild({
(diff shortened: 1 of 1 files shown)


lcd047:
This is not needed.  Please look at [this line](https://github.com/scrooloose/syntastic/blob/84add5b7f9a84dcc955d463a7dab639aa5bd031d/syntax_checkers/rst/rst2pseudoxml.vim#L21), then [this](https://github.com/scrooloose/syntastic/blob/84add5b7f9a84dcc955d463a7dab639aa5bd031d/syntax_checkers/rst/rst2pseudoxml.vim#L57), then read the [guide](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#isavailablecallback).

Can you please give an example of what you did, what did you expected to happen, and what happened instead?

mmv:
The code I proposed is wrong but the current behavior isn't right either: if you're running on Windows and `executable("rst2pseudoxml.py")` yields true, that doesn't mean you can just execute the file; you'll have to execute the python interpreter instead.

What ends up happening is that syntastic thinks it can execute the file and will spit errors whenever you attempt to save a `.rst` file.

lcd047:
If `executable("rst2pseudoxml.py")` returns true but `rst2pseudoxml.py` can't be run directly this is either a bug in Vim, or a config problem on your OS. _shrug_

lcd047:
Actually,  "[o]n MS-DOS and MS-Windows [`executable()`] only checks if the file exists and is not a directory, not if it's really executable.  Great, so how does one check if `foo` really is executable on Windows?

lcd047:
Ok, please try ebadf9a.

Issue #1336, opened 2015-02-24
Erlang Checker - Support for Makefiles etc


robashton:
I note that the code for the Erlang checker used to be quite similar to the vimerl one and they've diverged.

I'm not much of a fan of the "gather every folder you can find" approach taken in the Syntastic checker as it causes bugs and slowness so while I used to use it I've opted to improve vimerl's syntax checker and switch to that for now.

That said I'd like to keep Syntastic improving because it's a great project, so I'm wondering what the best way of getting this into Syntastic would be.

https://github.com/jimenezrick/vimerl/pull/80

I could create another syntax checker and call it "vimerl or Makefile aware", or just replace the one in Syntastic as it currently stands with the one in Vimerl (it doesn't seem as though there are any functional differences beyond the kitchen sink approach to compilation).

Thoughts? Happy to do the legwork but I don't really wanna make a PR that's gonna get rejected outright!

lcd047:
> That said I'd like to keep Syntastic improving because it's a great project, so I'm wondering what the best way of getting this into Syntastic would be.

Well you can explain me what problem are you trying to solve and how (keeping in mind that to me Erlang was a [Danish mathematician](http://en.wikipedia.org/wiki/Agner_Krarup_Erlang) rather than a programming language), or you can fight over it with @russss and @ahmadsherif, who were the last people to touch that piece of code.

robashton:
Hey,

Essentially - a lot of us are moving away from rebar anyway, and the current code not only is a bit rebar dependent but it's approach to finding includes/ebins is a tad indiscriminate - not only slow but also incorrect (erlc just flat out refuses to find some hrls)

The current vimerl is a bit similar (even more tied to rebar) but I've just submitted a PR to have fallback behaviour when a known "make" system is under use - finding the root directory properly and adding deps/etc based on that.

Adding this to the current compiler for syntastic is a bit unpleasant because of how it currently employ a scattergun approach to Building up options for erlc. So I was wondering whether to simply replace it with Vimerl's checker or adding it as an extra one.

As it currently stands, this checker does not work with our rebarless projects.

robashton:
Scuse lousy typing, I'm on my phone!

robashton:
(Vimerl's checker currently works with both kinds of project and it looks like they were originally the same - presumably somebody copied from somebody at some point - I can't see anything that the syntastic one does that the Vimerl one doesn't)

russss:
[...]

PR #1363, opened 2015-03-18
Erlang Syntax Checker that works with or without Rebar

robashton:
As promised in #1336 

This is the syntax checker I've written for vimerl, it has sensible default behaviour around rebar.config and if no rebar is available falls back to scanning for the default app structure (including in deps).

If anything is missing/broken then the best course of action is to update vimerl *and* this one - it looks like I have the latest fork of vimerl and the original isn't being updated so I guess I've got ownership for now.


lcd047:
Merged, thank you.

PR #1368, opened 2015-03-23
python: Read python source as bytes

myme:
Syntastic unsets LC_ALL when running commands, this leads to
`compile.py` dying horribly when encountering non-ascii characters in
source code running under Python 3.

Consuming the input source as bytes and passing this directly to
`compile` solves this:

### DIFF
--- a/syntax_checkers/python/compile.py
+++ b/syntax_checkers/python/compile.py
@@ -8,6 +8,6 @@ if len(argv) != 2:
     exit(1)
 
 try:
-    compile(open(argv[1]).read(), argv[1], 'exec', 0, 1)
+    compile(open(argv[1], 'rb').read(), argv[1], 'exec', 0, 1)
 except SyntaxError as err:
     print('%s:%s:%s: %s' % (err.filename, err.lineno, err.offset, err.msg))


lcd047:
Well, the meaning of the source shouldn't depend on environment variables.
```sh
$ cat reprod.py 
#! /usr/bin/env python
    a = 'ราชอาณาจักรไทย'

$ python -V
Python 2.7.9

$ LC_ALL= LC_CTYPE= LANG= TERM=dumb python reprod.py 
  File "reprod.py", line 2
SyntaxError: Non-ASCII character '\xe0' in file reprod.py on line 2, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```
If you do it the PEP 263 way, it works as expected:
```sh
$ cat reprod2.py 
#! /usr/bin/env python
# -*- coding: UTF-8 -*-
    a = 'ราชอาณาจักรไทย'

$ LC_ALL= LC_CTYPE= LANG= TERM=dumb python reprod2.py 
  File "reprod2.py", line 3
    a = 'ราชอาณาจักรไทย'
    ^
IndentationError: unexpected indent

$ python3 -V
Python 3.4.3

$ LC_ALL= LC_CTYPE= LANG= TERM=dumb python3 reprod2.py 
  File "reprod2.py", line 3
    a = '\u0e23\u0e32\u0e0a\u0e2d\u0e32\u0e13\u0e32\u0e08\u0e31\u0e01\u0e23\u0e44\u0e17\u0e22'
    ^
IndentationError: unexpected indent
```
It also works with syntastic:
```sh
$ LC_ALL= LC_CTYPE= LANG= TERM=dumb python compile.py reprod2.py 
reprod2.py:3:4: unexpected indent

$ LC_ALL= LC_CTYPE= LANG= TERM=dumb python3 compile.py reprod2.py 
reprod2.py:3:4: unexpected indent
```
That said, reading the sources in binary mode should be harmless, provided that it still works on Windows.  Does it?

lcd047:
Actually the last test is wrong, I was using a modified `compile.py`.  With the right script:
```sh
$  LC_ALL= LC_CTYPE=UTF-8 LANG= TERM=dumb python3 compile.py reprod2.py 
Traceback (most recent call last):
  File "compile.py", line 11, in <module>
    compile(open(argv[1]).read(), argv[1], 'exec', 0, 1)
  File "/usr/lib/python3.4/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe0 in position 56: ordinal not in range(128)
```
However:
```sh
$ LC_ALL= LC_CTYPE=en_US.UTF-8 LANG= TERM=dumb python compile.py reprod2.py 
reprod2.py:3:4: unexpected indent

$ LC_ALL= LC_CTYPE=en_US.UTF-8 LANG= TERM=dumb python3 compile.py reprod2.py 
reprod2.py:3:4: unexpected indent
```
So, to sum it up:

* checking files without `coding` against Python 2 doesn't make anything crash, but the results are less than useful;
* you can avoid the crashes by setting `LC_CTYPE` (the point of clearing `LC_ALL` and `LANG` is that syntastic needs to parse the output of the checkers, and localisation affects the format of said output);
* you can also avoid the crash by reading the files in binary mode, but this has to work on Windows with all major incarnations of Python;
* a similar change is needed for `codec.py`.

myme:
For some reason I have overlooked that my LC_CTYPE was set to something strange. I guess there's not really an issue to solve when the cause is a misconfigured system/environment. Feel free to disregard this patch, especially if it will cause trouble with the Windows test matrix. I'll close this and leave it up to you to re-open/merge if you feel that this patch has any other value.

PR #1374, opened 2015-03-31
check for puppet-lint not puppet

eikenb:
Puppetlint checker only runs if it finds the puppet executable which it
doesn't need or use. Change it to check for puppet-lint which is what it
does use.

### DIFF
--- a/syntax_checkers/puppet/puppetlint.vim
+++ b/syntax_checkers/puppet/puppetlint.vim
@@ -19,9 +19,9 @@ let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_puppet_puppetlint_IsAvailable() dict
-    call self.log("executable('puppet') = " . executable('puppet') . ', ' .
+    call self.log("executable('puppet-lint') = " . executable('puppet-lint') . ', ' .
         \ 'executable(' . string(self.getExec()) . ') = ' . executable(self.getExec()))
-    if !executable('puppet') || !executable(self.getExec())
+    if !executable('puppet-lint') || !executable(self.getExec())
         return 0
     endif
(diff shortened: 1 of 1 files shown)


lcd047:
1. Puppet-lint does need puppet to be installed.  If puppet is not installed, puppet-lint runs but it doesn't find all errors.
2. Syntastic does check for puppet-lint executable: it's the `executable(self.getExec())` in [this line](https://github.com/scrooloose/syntastic/blob/2337c65b37c94a4b1eb7b10059d911e7dd727cc2/syntax_checkers/puppet/puppetlint.vim#L24).

Sorry, but no.

eikenb:
Sorry I missed that getExec(), I should have caught that. 

I don't believe you are correct about puppet-lint requiring puppet to be installed to work. No where does it mention puppet as a requirement for use or development. None of the puppet-lint packages even suggest puppet needs to be installed, and the puppet-lint documentation lists the requirement and there is no mention of puppet.

lcd047:
> I don't believe you are correct about puppet-lint requiring puppet to be installed to work.

It certainly used to be true when I looked at it to fix the syntastic checker, but that was more than one year ago.  I think it used to be like this: if you installed `puppet-lint` with `gem install puppet-lint`, a number of module dependencies were not installed.  The modules were installed however if you also installed `puppet`.  Maybe the list of dependencies has been fixed in the mean time.  It would be useful to check with the developers.

rodjek:
puppet-lint hasn't required Puppet since v0.2.0 (mid 2012)

lcd047:
Ok, please try ebfba7e.

eikenb:
Works like a charm. Thanks.

PR #1378, opened 2015-04-06
map 'pe-puppet' to 'puppet' in register.vim

dudemcbacon:
In some cases vim will set the filetype of *.pp files to 'pe-puppet' instead of 'puppet'. There is no functional difference so lets map 'pe-puppet' to 'puppet' so we can use the same default checkers.

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -105,6 +105,7 @@ let s:_DEFAULT_FILETYPE_MAP = {
         \ 'litcoffee': 'coffee',
         \ 'mail': 'text',
         \ 'mkd': 'markdown',
+        \ 'pe-puppet': 'puppet',
         \ 'sgml': 'docbk',
         \ 'sgmllnx': 'docbk',
     \ }


lcd047:
This is not needed.  The easy way to achieve the same effect is to use `syntastic_filetype_map`:
```vim
let g:syntastic_filetype_map = { 'pe-puppet': 'puppet' }
```
On a more general note: it's amazing how much time reading the manual can save you.

dudemcbacon:
Ahh, I missed that, thank you.

Out of curiosity, why wouldn't we want this to be a default for syntastic? It would certainly save people the headache of having to figure out why their puppet files are no longer getting their syntax checked.

lcd047:
I'm not strongly opposed to it, the main disadvantage is that a hard-coded mapping cannot be undone by users.  What's the difference between `pe-puppet` and `puppet` filetypes anyway?

dudemcbacon:
They are functionally the same. The difference lies in how your puppet infrastructure is set up. For example, if you are using Puppet Enterprise, the packages for RHEL/CentOS will place a file at /usr/share/vim/vimfiles/ftdetect/pe-puppet.vim which sets the following:

```
" detect puppet filetype
au BufRead,BufNewFile *.pp              set filetype=pe-puppet
```

If you're using open-source Puppet, the packages will place a file at /usr/share/vim/vimfiles/ftdetect/puppet.vim which contains:

```
" detect puppet filetype
au BufRead,BufNewFile *.pp              set filetype=puppet
```

There is no syntactical difference between the files just inconsistencies with their naming in vim.

lcd047:
Ok, it makes sense to merge it after all.  Done in a17db15, thank you.

PR #1399, opened 2015-04-24
Ignore warning on checking rubocop's version

odiak:
Depends on installed ruby version, warning like below are appeared and syntastic cannot recognize rubocop's version.

```
$ rubocop --version
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.6-compliant syntax, but you are running 2.1.0.
0.30.0
```

This change fixes it.

### DIFF
--- a/syntax_checkers/ruby/rubocop.vim
+++ b/syntax_checkers/ruby/rubocop.vim
@@ -22,7 +22,7 @@ function! SyntaxCheckers_ruby_rubocop_IsAvailable() dict
     if !executable(self.getExec())
         return 0
     endif
-    return syntastic#util#versionIsAtLeast(self.getVersion(), [0, 12, 0])
+    return syntastic#util#versionIsAtLeast(self.getVersion(self.getExecEscaped() . ' --version 2>' . syntastic#util#DevNull()), [0, 12, 0])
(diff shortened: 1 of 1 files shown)


lcd047:
As I pointed out elsewhere, this is wrong.  First, `rubocop --version` shouldn't issue warnings to begin with, and when it does, it isn't syntastic's business to decide whether these warnings are survivable or not.  Then, killing _all_ warnings is not a solution, because this might hide other, less harmless problems.  The real solution is to fix your installation of ruby / rubocop so that `rubocop --version` doesn't print warnings.

But if you absolutely _must_ coerce syntastic to run with a broken configuration, you can write a wrapper around rubocop and point `g:syntastic_ruby_rubocop_exec` to it.  The wrapper might look something like this:
```sh
#! /bin/sh
if [ $# -eq 1 -a x"$1" = x--version ]; then
    rubocop "$@" 2>&1 | \
        sed \
            -e '/warning: parser\/current is loading parser\/ruby[0-9]*, which recognizes/d' \
            -e '/warning: [0-9.]*-compliant syntax, but you are running [0-9]/'                                                
else 
    exec rubocop "$@"
fi
```

PR #1400, opened 2015-04-24
Update url for tidy-html5

rstacruz:
They moved to @htacg, the "HTML Tidy Advocacy Community Group".


lcd047:
Which is why you should always look at the [wiki](https://github.com/scrooloose/syntastic/wiki/HTML%3A---tidy) (or, in this particular case, the [FAQ](https://github.com/scrooloose/syntastic#faqhtml5)) for user-level information.  I committed this instead: 9af644d.

PR #1401, opened 2015-04-25
First stab at YARD checker

turboladen:
This is for a checker that parses the output from [YARD](http://yardoc.org)'s `yard stats` command, which gives syntax warnings for its markup. It's a Ruby tool and I based it off of the Rubocop checker. I peeked at the guide and I think this should all be in line with that. I'm new to vim's `errorformat`, so suggestions are welcome.

### DIFF
--- a/syntax_checkers/ruby/yard.vim
+++ b/syntax_checkers/ruby/yard.vim
@@ -1,0 +1,58 @@
+"============================================================================
+"File:        yard.vim
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Steve Loveless
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"
+"============================================================================
+
+if exists('g:loaded_syntastic_ruby_yard_checker')
+    finish
+endif
+let g:loaded_syntastic_ruby_yard_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_ruby_yard_IsAvailable() dict
+    if !executable(self.getExec())
+        return 0
+    endif
+    return syntastic#util#versionIsAtLeast(self.getVersion(), [0, 7, 0])
+endfunction
+
+function! SyntaxCheckers_ruby_yard_GetLocList() dict
+    let makeprg = self.makeprgBuild({ 'args': 'stats' })
+
+    " Lines that contain the error info
+    let errorformat = '%W[warn]: %m,%Z    in file `%f'' near line %l,'
+    let errorformat .= '%E[error]: %m,%Z    in file `%f'' near line %l'
+
+    let loclist = SyntasticMake({
+        \ 'makeprg': makeprg,
+        \ 'errorformat': errorformat})
+
+    " Convert Yard severities to error types recognized by syntastic
+    for e in loclist
+        if e['type'] ==# 'e'
+            let e['type'] = 'E'
+        elseif e['type'] ==# 'w'
+            let e['type'] = 'W'
+        endif
+    endfor
+
+    return loclist
+endfunction
+
+call g:SyntasticRegistry.CreateAndRegisterChecker({
+    \ 'filetype': 'ruby',
+    \ 'name': 'yard'})
+
+let &cpo = s:save_cpo
+unlet s:save_cpo
+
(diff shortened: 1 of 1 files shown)


syntax_checkers/ruby/yard.vim
```diff
@@ -0,0 +1,58 @@
...
+        \ 'errorformat': errorformat})
+
+    " Convert Yard severities to error types recognized by syntastic
+    for e in loclist
+        if e['type'] ==# 'e'
+            let e['type'] = 'E'
+        elseif e['type'] ==# 'w'
+            let e['type'] = 'W'
+        endif
+    endfor
```
turboladen: Hmm... I'm kinda thinking I don't need this anymore. ...had been trying to match `[warn]` with `[%tarn]` but wasn't getting any results.

syntax_checkers/ruby/yard.vim
```diff
@@ -0,0 +1,58 @@
...
+    endif
+    return syntastic#util#versionIsAtLeast(self.getVersion(), [0, 7, 0])
+endfunction
+
+function! SyntaxCheckers_ruby_yard_GetLocList() dict
+    let makeprg = self.makeprgBuild({ 'args': 'stats' })
+
+    " Lines that contain the error info
+    let errorformat = '%W[warn]: %m,%Z    in file `%f'' near line %l,'
+    let errorformat .= '%E[error]: %m,%Z    in file `%f'' near line %l'
```
turboladen: Do you prefer a single, multi-line statement to this?

lcd047:
> This is for a checker that parses the output from [YARD](http://yardoc.org)'s `yard stats` command

Thank you.

> Hmm... I'm kinda thinking I don't need this anymore.

Yes, `type` is case-insensitive.

> Do you prefer a single, multi-line statement to this?

Formatting is a detail I can take care of; I'd prefer an `errorformat` that actually works. :) It's always a good idea to check the format of error messages in the sources.  Sadly, there are many cases your `errorformat` doesn't cover yet.

turboladen:
> Formatting is a detail I can take care of; I'd prefer an errorformat that actually works. :) 

I can't tell--are you suggesting that this `errorformat` doesn't work? It's seeming to catch all of the output from yard that I'm running into.

> It's always a good idea to check the format of error messages in the checker's sources.

That's a good idea--I didn't think of that. That'd be great to have in the guide for adding style checkers.

> Sadly, there are many cases your `errorformat` doesn't cover yet.

I can't tell--are you asking me add all of the cases or just pointing out the deficit?

lcd047:
> I can't tell--are you suggesting that this `errorformat` doesn't work?

Yes, that's what I'm saying.  Syntax errors in tags f.i. are not recognized:
```
$ yard stats test.rb
[warn]: Unknown tag @re_turn in file `test.rb` near line 121
Files:           1
Modules:         1 (    1 undocumented)
Classes:         1 (    1 undocumented)
Constants:       0 (    0 undocumented)
Methods:        12 (   10 undocumented)
 14.29% documented
```
Looking at the sources, there are many other problems: filenames are sometimes quoted as <code>\`...\`</code> and sometimes as <code>\`...'</code>, `in file ...` sometimes becomes `file ...`, and so on.  It's to be expected, most applications that don't have a centralized log formatter are prone to this mess.

> I can't tell--are you asking me add all of the cases or just pointing out the deficit?

Well, let's see.  The checker works for you, but it definitely won't work for everybody (I wasn't trying _that_ hard to break it).  An year from now, you'd have probably moved on from using it altogether, but I'd still be stuck with supporting the checker in syntastic.  You're using it enough to want to write a syntastic checker for it, while I have essentially zero interest in that entire class of underlying ideas.  You're fluent in Ruby, while I can just almost my way around on a  good day.  Who would be in the best position to do it?

Rummaging through the sources, not _all_ log formats seem relevant to `stats`.  If you'd compile a set of representative error samples, I could write the corresponding `eroformat` patterns for them (I've been doing that for a _long_ time now ;)).  If you feel particularly generous with your time, you might even try to get the problem fixed upstream.

turboladen:
Forget it man. It's not worth dealing with your passive aggressiveness. I ask you simple questions to clarify what you're trying to communicate via your insinuations and you respond with more passive aggressiveness/dickery. If you're that pissed off about maintaining this then it doesn't sound like it's worth your time, and I have no interest in collaborating with assholes.

lcd047:
> Forget it man. It's not worth dealing with your passive aggressiveness. I ask you simple questions to clarify what you're trying to communicate via your insinuations and you respond with more passive aggressiveness/dickery.

WTF?  I asked you to compile a set of representative error samples based on the sources, and I'll do the rest.  Ok, whatever, have a nice life finding people who aren't passive aggressive assholes to collaborating with.

PR #1405, opened 2015-05-02
Improvements/fixes for auto-location-list handling

blueyed:
Especially / only(?) for `g:syntastic_auto_loc_list = 1`.


blueyed:
Hmm - but the owner would still be (or could be assumed to be) the window/buffer(?) that it was run from?

blueyed:
JFI: only https://github.com/blueyed/syntastic/commit/ce9db766846fa2a493269296e52b0d690f3fa92b is now remaining from this PR.

PR #1406, opened 2015-05-02
Fix SyntasticLoclist.isNewerThan (Neovim, but in general)

blueyed:
`reltime()` returns a list of integers, where the second one might be
negative, e.g. `[0, -200945546]` (for 4.094022) seconds.

Using `syntastic#util#compareLexi` for comparison would return 0 in that
case, resulting in the notifiers not being called.

Since the second entry from `reltime()` is not always negative (or the first one might have advanced) this caused Syntastic to sometimes just not opening the location list with `g:syntastic_auto_loc_list = 1` and caused some other weird / not easily to reproduce issues probably.

### DIFF
--- a/plugin/syntastic/loclist.vim
+++ b/plugin/syntastic/loclist.vim
@@ -59,9 +59,16 @@ endfunction " }}}2
 function! g:SyntasticLoclist.isNewerThan(stamp) abort " {{{2
     if !exists('self._stamp')
         let self._stamp = []
+    endif
+    if a:stamp == []
+        return 1
+    endif
+    if self._stamp == []
         return 0
     endif
-    return syntastic#util#compareLexi(self._stamp, a:stamp) > 0
+    let [a_s, a_ms] = split(split(reltimestr(reltime(self._stamp)))[0], '\.')
+    let [b_s, b_ms] = split(split(reltimestr(reltime(a:stamp)))[0], '\.')
+    return a_s > b_s || (a_s == b_s && a_ms > b_ms)
 endfunction " }}}2
 
 function! g:SyntasticLoclist.copyRaw() abort " {{{2


blueyed:
That seems to be specific to Neovim..! I have not seen negative numbers with Vim itself.

ZyX-I:
String comparison is a bad idea because `"14.0" < "4.0"`. Note that documentation explicitly says that format returned by `reltime()` is system-specific, so code relying on any exact format is already broken. `reltimestr()` format is specified as `[spaces]{seconds}.{microseconds}` without an exact size of `{seconds}` part, so to compare you should use something like

    let [seconds, microseconds]=[[0, 0], [0, 0]]
    let [seconds[0], microseconds[0]] = split(split(reltimestr(self._stamp))[0], '\V\c.')
    let [seconds[1], microseconds[1]] = split(split(reltimestr(a:stamp))[0], '\V\c.')
    let seclen = max(map(seconds[:], 'len(v:val)'))
    call map(microseconds, 'printf("%0*s%06s", seclen, seconds[v:key], v:val)')
    return microseconds[0] ># microseconds[1]

blueyed:
@ZyX-I 
Thanks, you are right of course.

I'd prefer the following though:

    function! g:SyntasticLoclist.isNewerThan(stamp) abort " {{{2
        if !exists('self._stamp')
            let self._stamp = []
            return 0
        endif
        if a:stamp == []
            return 1
        endif
        let [a_s, a_ms] = split(split(reltimestr(reltime(self._stamp)))[0], '\.')
        let [b_s, b_ms] = split(split(reltimestr(reltime(a:stamp)))[0], '\.')
        return a_s > b_s || (a_s == b_s && a_ms > b_ms)
    endfunction " }}}2

blueyed:
Amended the commit.
Please let me know if I'd have missed something again.

ZyX-I:
>     let [a_s, a_ms] = split(split(reltimestr(reltime(self._stamp)))[0], '\.')
>     let [b_s, b_ms] = split(split(reltimestr(reltime(a:stamp)))[0], '\.')
>     return a_s > b_s || (a_s == b_s && a_ms > b_ms)

This is incorrect. You *must* use `printf` because otherwise you are doing string comparison on strings with different lengths. If not that you must compare string lengths. This way you are *also* having incorrect result for `14.0` and `4.0`: in the example `a_s` is `"14"`, `b_s` is `"4"` and both `a_s > b_s` and `a_s == b_s` will return zero resulting in the whole expression returning zero in place of one.

lcd047:
Thank you.  Please try commit 3a30321.

PR #1411, opened 2015-05-06
pylama: requires '--force' if the file does not end in .py

blueyed:
Ref: https://github.com/klen/pylama/pull/35

### DIFF
--- a/syntax_checkers/python/pylama.vim
+++ b/syntax_checkers/python/pylama.vim
@@ -27,7 +27,7 @@ function! SyntaxCheckers_python_pylama_GetHighlightRegex(item)
 endfunction
 
 function! SyntaxCheckers_python_pylama_GetLocList() dict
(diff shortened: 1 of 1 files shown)


lcd047:
This is not enough: option `--force` seems to have appeared in pylama version 4.0.

blueyed:
Oh.. this would require a version detection in Syntastic?

Then I'd wait for the outcome of https://github.com/klen/pylama/pull/35, and not add `--force` with Syntastic, if it would be fixed with a recent pylama.

lcd047:
> Oh.. this would require a version detection in Syntastic?

Yup.

lcd047:
There doesn't seem to be much interest in fixing this upstream, so I went ahead and did it anyway in 0470fff.

blueyed:
Thanks!

PR #1413, opened 2015-05-09
Add tentative ponyc checker

Earnestly:
This patch adds support for the new [ponyc](https://github.com/CausalityLtd/ponyc) programming language.

There are some issues which can be better handled but the current result
seems to work.

I have made comments about this issue but the use of the flag `-p /usr/lib/pony`
is Arch Linux specific at the moment which will hopefully be fixed soon
so that `ponyc` will look in this location by default. https://github.com/CausalityLtd/ponyc/issues/172

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -63,6 +63,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'php':           ['php', 'phpcs', 'phpmd'],
         \ 'po':            ['msgfmt'],
         \ 'pod':           ['podchecker'],
+        \ 'pony':          ['ponyc'],
         \ 'puppet':        ['puppet', 'puppetlint'],
         \ 'python':        ['python', 'flake8', 'pylint'],
         \ 'r':             [],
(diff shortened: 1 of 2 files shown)


lcd047:
The Pony compiler looks experimental to the point of its UI still being a moving target.  Perhaps your checker should be added to [pony-vim-syntax](https://github.com/dleonard0/pony-vim-syntax) instead?

Earnestly:
It is pretty new.  I would personally not merge this but essentially leave it
up as google fodder.

As for the syntax, I'm not sure how you would include standalone checkers for
syntastic in that manner so I hadn't thought of it.

lcd047:
> As for the syntax, I'm not sure how you would include standalone checkers for syntastic in that manner so I hadn't thought of it. Can you elaborate on that?

It's described in the [guide](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).  Basically all you need to do is add `pony` to `g:syntastic_extra_filetypes`.

The point of making it external is that people who care about both Pony and Vim will get `pony-vim-syntax` anyway, and there it will be maintained by somebody who actually follows the development of the language.

Earnestly:
That's not a bad idea.  Sorry I missed the external documentation in the guide, thanks for pointing it out.

Earnestly:
I'm going to close this for now.  It should be locatable by anyone looking for
it.  The language compiler is still very new and it'll be awhile before the
rough edges are ironed out.

This allows anyone with more vim skill to create something better.

I'll keep the external checker available for the syntax plugin though, thanks
for the suggestion.

PR #1414, opened 2015-05-09
Compile elixir rather than execute it.

### LINKED ISSUE #1141: Elixir checker executes code
The Elixir syntax checker executes any program it checks. For example:

``` elixir
# sleeper.exs
sleeper = fn -> :timer.sleep(4_000) end
sleeper.()
```

It takes 4 seconds to save this file because the `sleeper` function is executed.

This method of checking is dangerous; for example, suppose it were a script to delete files? Also it's annoying; if you accidentally create infinite recursion, Vim will hang.

Can we check syntax without executing the code?

(The answer may be "only in a limited way": see [this thread](https://groups.google.com/d/msg/elixir-lang-talk/B29noPHvQ-8/9JvSGPop7n0J))

ericlathrop:
Use `elixirc` to compile elixir files for syntax checks rather than
executing them with `elixir`.

### DIFF
--- a/syntax_checkers/elixir/elixir.vim
+++ b/syntax_checkers/elixir/elixir.vim
@@ -21,20 +21,14 @@ set cpo&vim
 " TODO: we should probably split this into separate checkers
 function! SyntaxCheckers_elixir_elixir_IsAvailable() dict
     call self.log(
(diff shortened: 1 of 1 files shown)


lcd047:
A similar change has been rejected in #1343.  Sorry, but no.

PR #1420, opened 2015-05-19
Add support for flow server

shayne:
Add support to use the flow server (much faster) via the syntastic_javascript_flow_use_server var.

### DIFF
--- a/syntax_checkers/javascript/flow.vim
+++ b/syntax_checkers/javascript/flow.vim
@@ -26,8 +26,10 @@ function! SyntaxCheckers_javascript_flow_GetLocList() dict
         return []
     endif
 
+    let use_server = get(g:, 'syntastic_javascript_flow_use_server')
+
     let makeprg = self.makeprgBuild({
-        \ 'exe': self.getExecEscaped() . ' check',
+        \ 'exe_after': (use_server ? 'status' : 'check'),
         \ 'args_after': '--show-all-errors --json' })
 
     let errorformat =


lcd047:
It's already possible to to achieve that by setting `g:syntastic_javascript_flow_exe` to `/path/to/flow status` (cf. #1341).  This isn't the default because it needs the server to be started before running the checks.  Please add a note to the [wiki](https://github.com/scrooloose/syntastic/wiki/JavaScript%3A---flow) to that effect if you feel it would be useful.

shayne:
Thanks for pointing me to g:syntastic_javascript_flow_exe. When you say
this isn't the default because it needs the server started... that's not
the case at least with recent versions of flow. Issuing `flow check`
without the server running will actually kick off a server.

lcd047:
@shayne: `flow check` is the current behaviour.  The point is, issuing `flow status` won't start a server by itself, and it doesn't work if the server isn't running.  Or at least that used to be the case when I checked (which was about when #1341 was posted).

shayne:
For me "flow status" does in fact start the server...

`sweeney at sweeney-mbp1 in ~/Code/CirrusForums$ flow status
Flow server launched for /Users/sweeney/Code/CirrusForums
Spawned flow server (child pid=7203)
Logs will go to
/var/folders/5p/bpr5yyxj1vxf31tcm50jz1khpsfv0q/T/flow/zSUserszSsweeneyzSCodezSCirrusForums.log
The flow server will be ready in a moment.
Error: flow server is busyzing. If it was just started this can take some
time. Retrying... |`

lcd047:
Ok, then this should work: 981196f.

PR #1422, opened 2015-05-20
replacing gcc to standard system's cc compiler

paranormal:


lcd047:
This is not needed.  You can achieve the same effect without changing the code:
```vim
let g:syntastic_c_compiler = 'cc'
let g:syntastic_c_compiler_options = '-std=c99'
```
The checker will be still named `gcc`, but it will be running `cc` for checks.

paranormal:
I understand that, buy why anyone need to do this? I strongly think that this is must be default configuration, Why would anyone redefine gcc to cc in case not having gcc? Looks like absurd.

This is needed if we'd like to comply with standards and provide working configuration by default.

paranormal:
I mean gcc users wouldn't notice this patch, when for others it mean less configuration.

And it seems I've forgotten to provide patch for plugin/syntastic/registry.vim.

Let me know, do I have any chance to have this committed? I'll make pull request again then.

lcd047:
> I understand that, buy why anyone need to do this?

Because the `errorformat` has been taylored and tested for `gcc`, and not for any of the many `cc` compilers available out there?  If you set the two variables as I pointed out above the checker may or may not work, depending on what your `cc` happens to be.  Chances are it will work, but this hasn't been tested.  In contrast, the checker has been tested extensively with `gcc` and `clang`.

> Let me know, do I have any chance to have this committed?

Sorry, but no.

paranormal:
Pity to hear this, the few with icc would definitely know what they do, since they must install it at least, commercial unix users may be affected though, but they are affected with their compilers anyway.
Those categories of users fit for editing config file.

But ok I'll live with this :) Thank you for marvellous tool!

PR #1429, opened 2015-05-26
Add support for Vera++ C++ checker

Phyks:
This pull request adds support for Vera++ C++ style checker, using its ability to export to cppstyle format, and then based on the checkstyle checker.

Let me know if you have any comments about eventual bad practice in this checker.


lcd047:
Merged, thank you.

Issue #970, opened 2014-01-29
New feature: options 'args_before' and 'args_after' for makeprgBuild()


lcd047:
This is a bit hard to explain, but you'll probably like it.  Basically, the `args` option to `makeprgBuild()` (see the [wiki](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide)) has been split into `args_before`, `args`, and `args_after`.  The options still expand to `<args_before> <args> <args_after>` in the resulting `makeprg`, but only the middle term `<args>` can be overridden by `g:syntastic_<filetype>_<checker>_args`.  The other two terms can't be changed at runtime, and are always included in the result.  (In fact, there are `*_before` and `*_after` variants for all options, not just `args`, but you'll probably only ever care about `args`, and perhaps occasionally about `post_args`).

The point of all this is to keep immutable the options that tell the checker to behave in the way expected by syntastic, while still allowing user to add other options.  Take for example `pylint`.  You can now put

``` vim
let makeprg = self.makeprgBuild({
        \ 'args_after': (s:pylint_new ? '-f text --msg-template="{path}:{line}:{column}:{C}: [{symbol}] {msg}" -r n' : '-f parseable -r n -i y') })
```

in the checker, and the user will be able to set, say,

``` vim
let g:syntastic_python_pylint_args = '-d line-too-long'
```

and not worry about output formats, versions, and other details that are only relevant to syntastic.

Please update your `vimrc`s, and test this thoroughly.  The change was not complicated, but a lot of code has moved around again, and my choice of what options should be fixed might not have been always the best.  Commit 46a2181.

jmjohnson:
Howdy, just thought I'd leave a note that the inability to change `args_after` has caused me a little grief while trying to implement my company's custom rubocop formatter.

I want to pass a single `--format` flag to rubocop but the immutable `args_after` makes it impossible to use something convenient like `let g:syntastic_ruby_rubocop_args_after = ""`. ([source code](https://github.com/vim-syntastic/syntastic/blob/96cc251075f3f9d290c5afd4adf1b64c1542d469/syntax_checkers/ruby/rubocop.vim#L29)).

I'm sure I'll be able to hack something together (perhaps implement my own syntax checker?), but just thought I'd register my use case for making `args_after` configurable. 😃

lcd047:
> the inability to change `args_after` has caused me a little grief while trying to implement my company's custom rubocop formatter.

[...]

PR #1439, opened 2015-06-04
Crystal external checker

veelenga:
To make people aware...

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -66,7 +66,7 @@ YAML, z80, Zope page templates, and zsh.  See the [wiki][3] for details about
 the corresponding supported checkers.
 
 A number of third-party Vim plugins also provide checkers for syntastic,
-for example: [omnisharp-vim][25], [rust.vim][12], [syntastic-extras][26],
+for example: [omnisharp-vim][25], [rust.vim][12], [crystal.vim][29], [syntastic-extras][26],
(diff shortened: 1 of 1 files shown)


lcd047:
Thank you.  I'd suggest changing `args` to `args_after` [here](https://github.com/rhysd/vim-crystal/blob/master/syntax_checkers/crystal/crystal.vim#L18), but otherwise, sure, merged.

veelenga:
Found an explanation [here](https://github.com/scrooloose/syntastic/issues/970). Sure, I will! Thanks for suggestion!

lcd047:
On second thought, this looks more appropriate:
```vim
        \ 'exe_after': 'run',
        \ 'args_after': '--no-build --no-color' })
```

veelenga:
Original command looks like:
```sh
crystal run --no-build --no-color source.cr
```

Does your suggestion mean that user can pass parameter between "run" and "--no-build --no-color..." ?  Using appropriate variable:

```viml
let g:syntastic_crystal_crystal_args = '--release'
```

And after the command will look like:

```sh
crystal run --release --no-build --no-color source.cr
```

lcd047:
Yes, that's pretty much the point.

veelenga:
Awesome. Thanks!

PR #1444, opened 2015-06-09
javascript/flow: use 2nd location instead of first if available

seanhess:
Right now, if you have this code: 

    type User = {name: string, age: number}

    // notice I forgot the age!
    var user:User = {name:"hello"} 

Flow reports an error that gives both the line number of the type, and of the object. Right now we're using the line number of the type, which is very confusing. 

I added another errorformat entry that will use the 2nd location when available so it correctly points to the object missing the property. 

----

Example: If we have this error from the preprocessor: 

    /Users/seanhess/projects/serials/web/app/model/source.js:20:22: property hidden Property not found in object literal (/Users/seanhess/projects/serials/web/app/model/source.js:94:10,106:3)

It will now result in this output in the location list: 

    web/app/model/source.js|94 col 10 error| property hidden Property not found in object literal [javascript/flow]

### DIFF
--- a/syntax_checkers/javascript/flow.vim
+++ b/syntax_checkers/javascript/flow.vim
@@ -38,6 +38,7 @@ function! SyntaxCheckers_javascript_flow_GetLocList() dict
         \ 'args_after': '--show-all-errors --json' })
 
     let errorformat =
+        \ '%.%#: %m (%f:%l:%c\,%.%#),' .
         \ '%f:%l:%c:%n: %m,' .
         \ '%f:%l:%c: %m'
 


seanhess:
I'm happy to clean this up if you'd prefer a different approach. We could hack the preprocessor, but this is definitely an issue in real projects. 

@faceleg @lcd047 not sure if you guys are the right ones to review. Thanks for your help on the stackoverflow issue lcd!

lcd047:
Your patch is wrong because it _always_ gives priority to the second location.  Take this piece of code:
```js
/* @flow */

function total(numbers: Array<number>) {
  var result = 0;
  for (var i = 0; i < numbers.length; i++) {
    result += numbers[i];
  }
  return result;
}

total([1, 'e', 3, 'Hello' + ' hh ' ])
```
Without your patch, syntastic points you to the last line.  With your patch however the error is located at the function definition, which is wrong.

If this is to be fixed, it should be addresses when parsing the JSON structure.  Trying to fix it after the fact is, again, wrong.  Now, the JSON structure looks like this for your sample code:
```json
{
   "passed" : false,
   "errors" : [
      {
         "message" : [
            {
               "end" : 39,
               "path" : "/home/lcd047/tmp/test.js",
               "start" : 13,
               "code" : 0,
               "endline" : 3,
               "line" : 3,
               "descr" : "property age\nProperty not found in"
            },
            {
               "descr" : "object literal",
               "endline" : 6,
               "line" : 6,
               "end" : 30,
               "start" : 17,
               "code" : 0,
               "path" : "/home/lcd047/tmp/test.js"
            }
         ]
      }
   ],
   "version" : " May 12 2015 22:01:53"
}
```
It looks like this for mine:
```json
{
   "errors" : [
      {
         "message" : [
            {
               "descr" : "function call\nError:",
               "start" : 1,
               "end" : 37,
               "endline" : 11,
               "path" : "/home/daia/tmp/test2.js",
               "line" : 11,
               "code" : 0
            },
            {
               "endline" : 11,
               "end" : 13,
               "start" : 11,
               "descr" : "string\nThis type is incompatible with",
               "code" : 0,
               "path" : "/home/daia/tmp/test2.js",
               "line" : 11
            },
            {
               "start" : 31,
               "descr" : "number",
               "end" : 36,
               "endline" : 3,
               "path" : "/home/daia/tmp/test2.js",
               "line" : 3,
               "code" : 0
            }
         ]
      },
      {
         "message" : [
            {
               "endline" : 11,
               "end" : 37,
               "descr" : "function call\nError:",
               "start" : 1,
               "code" : 0,
               "path" : "/home/daia/tmp/test2.js",
               "line" : 11
            },
            {
               "end" : 34,
               "endline" : 11,
               "start" : 19,
               "descr" : "string\nThis type is incompatible with",
               "code" : 0,
               "path" : "/home/daia/tmp/test2.js",
               "line" : 11
            },
            {
               "start" : 31,
               "descr" : "number",
               "end" : 36,
               "endline" : 3,
               "path" : "/home/daia/tmp/test2.js",
               "line" : 3,
               "code" : 0
            }
         ]
      }
   ],
   "version" : " May 12 2015 22:01:53",
   "passed" : false
}
```

lcd047:
Your patch is wrong because it _always_ gives priority to the second location.  Take this piece of code:
```js
/* @flow */

function total(numbers: Array<number>) {
  var result = 0;
  for (var i = 0; i < numbers.length; i++) {
    result += numbers[i];
  }
  return result;
}

total([1, 'e', 3, 'Hello' + ' hh ' ])
```
Without your patch, syntastic points you to the last line.  With your patch however the error is located at the function definition, which is wrong.

If this is to be fixed, it should be addresses when parsing the JSON structure.  Trying to fix it after the fact is, again, wrong.  Now, the JSON structure looks like this for your sample code:
```json
{
   "passed" : false,
   "errors" : [
      {
         "message" : [
            {
               "end" : 39,
               "path" : "/home/lcd047/tmp/test.js",
               "start" : 13,
               "code" : 0,
               "endline" : 3,
               "line" : 3,
               "descr" : "property age\nProperty not found in"
            },
            {
               "descr" : "object literal",
               "endline" : 6,
               "line" : 6,
               "end" : 30,
               "start" : 17,
               "code" : 0,
               "path" : "/home/lcd047/tmp/test.js"
            }
         ]
      }
   ],
   "version" : " May 12 2015 22:01:53"
}
```
It looks like this for mine:
```json
{
   "errors" : [
      {
         "message" : [
            {
               "descr" : "function call\nError:",
               "start" : 1,
               "end" : 37,
               "endline" : 11,
               "path" : "/home/lcd047/tmp/test2.js",
               "line" : 11,
               "code" : 0
            },
            {
               "endline" : 11,
               "end" : 13,
               "start" : 11,
               "descr" : "string\nThis type is incompatible with",
               "code" : 0,
               "path" : "/home/lcd047/tmp/test2.js",
               "line" : 11
            },
            {
               "start" : 31,
               "descr" : "number",
               "end" : 36,
               "endline" : 3,
               "path" : "/home/lcd047/tmp/test2.js",
               "line" : 3,
               "code" : 0
            }
         ]
      },
      {
         "message" : [
            {
               "endline" : 11,
               "end" : 37,
               "descr" : "function call\nError:",
               "start" : 1,
               "code" : 0,
               "path" : "/home/lcd047/tmp/test2.js",
               "line" : 11
            },
            {
               "end" : 34,
               "endline" : 11,
               "start" : 19,
               "descr" : "string\nThis type is incompatible with",
               "code" : 0,
               "path" : "/home/lcd047/tmp/test2.js",
               "line" : 11
            },
            {
               "start" : 31,
               "descr" : "number",
               "end" : 36,
               "endline" : 3,
               "path" : "/home/lcd047/tmp/test2.js",
               "line" : 3,
               "code" : 0
            }
         ]
      }
   ],
   "version" : " May 12 2015 22:01:53",
   "passed" : false
}
```
If you can come up with some algorithm to produce a line and a column that works in _all_ cases, please feel free to post it.  In the mean time, the above doesn't work, sorry.

seanhess:
The only thing I can think of is to default to the first message, like now, but if the first message matches specific things flip it around. 

So in this case, if we find `"Property not found in"` in the first error message `descr` then use the 2nd. I can't remember where else I've run in to this problem but it's been often enough that I don't think it's limited to object literals. I could implement this one and then make new PRs for other out-of-order results as I find them.

lcd047:
Actually, I'd say this should be fixed upstream.  Trying to deal in syntastic with the usability quirks of the various checkers is a losing battle, and this is particularly true for flow.  Facebook has a lot more manpower for introducing this kind of problems than I have to deal with them, ;) and this is only one checker out of ~180 (for now).

Have you tried posting a bug report upstream?  I would really prefer to keep this out of syntastic.

seanhess:
Yeah, I opened one on facebook/flow. https://github.com/facebook/flow/issues/501. Thanks!

lcd047:
Thank you.  Please let me know if this ends up with changes to the error format that require my intervention.

PR #1447, opened 2015-06-12
(Erlang) Apps should be used for include_lib too

robashton:
Just a small one - include_lib should work for multiple app projects as well and this PR does that.


lcd047:
Merged, thank you.

PR #1453, opened 2015-06-20
Add support for .nix files

gfxmonk:
This is a simple syntax checker for nix files (http://nixos.org/nix/). It uses `nix-instantiate --eval-only` to validate the current file's syntax.


lcd047:
Thank you.  Merged, with this patch on top: 98e79d8

PR #1454, opened 2015-06-22
use the sphinx "pseudoxml" builder as a syntax checker

bukzor:
This probably isn't 100% right, but it's a good starting point, and it works better for me than the existing rst checkers.

I believe this should be the preferred checker for python documentation, but I don't know how I would implement that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+*.pyc
 *.swp
 tags
 .DS_Store
--- a/README.markdown
+++ b/README.markdown
@@ -60,10 +60,10 @@ Haml, Haskell, Haxe, Handlebars, HSS, HTML, Java, JavaScript, JSON, JSX, LESS,
 Lex, Limbo, LISP, LLVM intermediate language, Lua, Markdown, MATLAB, Mercury,
 NASM, Nix, Objective-C, Objective-C++, OCaml, Perl, Perl POD, PHP, gettext
 Portable Object, OS X and iOS property lists, Puppet, Python, R, Racket, Relax
-NG, reStructuredText, RPM spec, Ruby, SASS/SCSS, Scala, Slim, SML, Tcl, TeX,
-Texinfo, Twig, TypeScript, Vala, Verilog, VHDL, VimL, xHtml, XML, XSLT, YACC,
-YAML, z80, Zope page templates, and zsh.  See the [wiki][3] for details about
-the corresponding supported checkers.
+NG, reStructuredText, RPM spec, Ruby, SASS/SCSS, Scala, Slim, SML, Sphinx, Tcl,
+TeX, Texinfo, Twig, TypeScript, Vala, Verilog, VHDL, VimL, xHtml, XML, XSLT,
+YACC, YAML, z80, Zope page templates, and zsh.  See the [wiki][3] for details
+about the corresponding supported checkers.
 
 A number of third-party Vim plugins also provide checkers for syntastic,
 for example: [omnisharp-vim][25], [rust.vim][12], [syntastic-extras][26],
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -69,7 +69,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'r':             [],
         \ 'racket':        ['racket'],
         \ 'rnc':           ['rnv'],
-        \ 'rst':           ['rst2pseudoxml'],
+        \ 'rst':           ['sphinx'],
         \ 'ruby':          ['mri'],
         \ 'sass':          ['sass'],
         \ 'scala':         ['fsc', 'scalac'],
--- a/syntax_checkers/rst/sphinx.vim
+++ b/syntax_checkers/rst/sphinx.vim
@@ -18,11 +18,18 @@ let g:loaded_syntastic_rst_sphinx_checker = 1
 let s:save_cpo = &cpo
 set cpo&vim
 
+let s:sphinx_cache_location = syntastic#util#tmpdir()
+lockvar s:sphinx_cache_location
+
+augroup syntastic
+    autocmd VimLeave * call syntastic#util#rmrf(s:sphinx_cache_location)
+augroup END
+
 function! SyntaxCheckers_rst_sphinx_GetLocList() dict
 
     let srcdir = syntastic#util#var('rst_sphinx_source_dir')
     if srcdir == ''
-        let config = findfile('conf.py', '.;')
+        let config = findfile('conf.py',  escape(expand('%:p:h', 1), ' ') . ';')
         if config == ''
             return []
         endif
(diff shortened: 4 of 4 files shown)


bukzor:
Current issues:

 1. The "." in makeprgBuild will likely break down if $PWD isn't the same directory as conf.py.
 2. There's no distinction between warning and error. This is nice-to-have imo.
 3. This plugin isn't automatically activating for me. I have to set `g:syntastic_rst_checkers = ['sphinx']` to make it work.

bukzor:
Yea... The first argument needs to be the first parent directory (of the file) that contains `conf.py`.

I'm not sure how to make that work. I suppose I'd have to write a VimL function for such?

lcd047:
Your 1. could be dealt with, by trying to locate `conf.py`.  2. is a minor annoyance, and 3. is to be expected, since not all checkers should be automatically activated.  However, generating unwanted output is a deal breaker, sorry.

I made an [attempt](https://github.com/scrooloose/syntastic/commit/ca6fffdb8e753df2d729bd34344d6bbaec7cc426) at this myself, a few month ago (branch [sphinx_checker](https://github.com/scrooloose/syntastic/tree/sphinx_checker)).  My patch adds a new builder named `dummy`, that does everything except write the final result.  The deal breaker for it was that the builder has to be activated in `conf.py` (IIRC).  Still, I think this is a cleaner solution than yours.

If somebody could check my patch for sanity, add an option to `sphinx` to activate it from the command line, and get it included upstream, I'd say this would be the way to go.

bukzor:
I've fixed my !#1 using the `findInParent` util.

I note that there's currently three other checkers that generate unwanted output, and it seems to be the entire reason util#rmrf exists. They are co/coco java/javac lisp/clisp and sass/sass.

I'll probably be using this as-is, although I'll give your branch a quick spin.

bukzor:
Please note that your dummy builder still doesn't avoid writing out the weird pickle file that sphinx uses.
The rmrf strategy handles this (relatively) cleanly though.

lcd047:
> I've fixed my !#1 using the  findInParent  util.

And somehow you missed to specify an `errorformat`?  Look, all that has been dealt with in my patch already.

> I note that there's currently four other checkers that generate unwanted output, and it seems to be the entire reason util#rmrf exists. They are co/coco java/javac lisp/clisp and sass/sass.

Yes.  `util#rmrf()` is a `rm -rf` waiting to happen. :) I'd still prefer to avoid using it if possible.

bukzor:
OK, yes, but where we can't avoid it, such as sass and sphinx, it's the only alternative.
I'll work on getting a `dummy` builder in upstream sphinx, but it will take a month at least, and won't help anyone that has slightly-older sphinx installed.

I submitted this patch with no `errorformat` because it worked without it.
That said, I'll merge yours.

I only run #rmrf` with the  temporary directory that I created with `util#tmpdir`.
That's as safe as it can get.

lcd047:
I cleaned that up, commit 3f5bab3.  It still needs logging (the mechanisms for logging were different at the time I wrote that), and better handling of initial errors (letting the user know about missing extensions and the like).

The main point of trying to merge this upstream is not to add the `dummy` builder (there's no problem shipping it with syntastic), but to patch `sphinx-builder` to allow adding extensions from the command line.  That's the only thing not handled in my patch.  Without it you have to add `syntastic` to the list of extensions in `conf.py`, which means you'd also have to take it out before making a release.  This is way too much of a pain to be acceptable.

As for using `pseudoxml`, that slow things down.  I suppose you could maintain it as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external) until `dummy` gets merged upstream, and make it a fallback option for people with older sphinxes after the merge.

bukzor:
I've merged your changes.
For your convenience, here's the diff between your branch and mine (after merging master) http://paste.pound-python.org/show/6cuY8553kjYM3k8DblSA/


A biggest difference between the two approaches is that your dummy writer doesn't catch this class of issues at all:

```
nail/home/buck/trees/theirs/requests/requests/api.py:docstring of requests.request:None: WARNING: py:obj reference target not found: float or tuple
/nail/home/buck/trees/theirs/requests/requests/api.py:docstring of requests.request:None: WARNING: py:obj reference target not found: bool
/nail/home/buck/trees/theirs/requests/docs/api.rst:127: WARNING: py:class reference target not found: HTTPError
/nail/home/buck/trees/theirs/requests/docs/api.rst:101: WARNING: py:class reference target not found: TooManyRedirects
/nail/home/buck/trees/theirs/requests/requests/sessions.py:docstring of requests.Session.request:None: WARNING: py:obj reference target not found: float or tuple
/nail/home/buck/trees/theirs/requests/requests/sessions.py:docstring of requests.Session.request:None: WARNING: py:obj reference target not found: bool
/nail/home/buck/trees/theirs/requests/requests/adapters.py:docstring of requests.adapters.HTTPAdapter:None: WARNING: py:obj reference target not found: int
/nail/home/buck/trees/theirs/requests/requests/adapters.py:docstring of requests.adapters.HTTPAdapter.send:None: WARNING: py:obj reference target not found: float or tuple
/nail/home/buck/trees/theirs/requests/docs/api.rst:127: WARNING: py:class reference target not found: HTTPError
/nail/home/buck/trees/theirs/requests/docs/api.rst:101: WARNING: py:class reference target not found: TooManyRedirects
/nail/home/buck/trees/theirs/requests/requests/sessions.py:docstring of requests.Session.request:None: WARNING: py:obj reference target not found: float or tuple
/nail/home/buck/trees/theirs/requests/requests/sessions.py:docstring of requests.Session.request:None: WARNING: py:obj reference target not found: bool
/nail/home/buck/trees/theirs/requests/requests/adapters.py:docstring of requests.adapters.HTTPAdapter:None: WARNING: py:obj reference target not found: int
/nail/home/buck/trees/theirs/requests/requests/adapters.py:docstring of requests.adapters.HTTPAdapter.send:None: WARNING: py:obj reference target not found: float or tuple
```

To correct this, change your stub of `.write` to a stub of `.write_doc`:

```#!diff
diff --git a/syntax_checkers/rst/syntastic.py b/syntax_checkers/rst/syntastic.py
index 198dae3..f67b9d8 100644
--- a/syntax_checkers/rst/syntastic.py
+++ b/syntax_checkers/rst/syntastic.py
@@ -25,7 +25,7 @@ class DummyBuilder(Builder):
     def prepare_writing(self, docnames):
         pass

-    def write(self, *ignored):
+    def write_doc(self, docname, doctree):
         pass

```

Since the main argument seems to be performance, I did five runs by each method on the requests index.rst: (after applying above patch to your dummy writer)

 # | pseudoxml | dummy |
--- | -------------- | ---------- |
1   | 3.929 *      | 4.440    |
2   | 4.125         | 3.565 * |
3   | 3.874         | 3.551    |
4   | 3.975         | 3.600    |
5   | 3.852         | 3.516    |

Median times are 3.929 and 3.565 respectively. The difference is 9%.
While the speedup is significant, it borders on subliminal.

If you actually wanted to make it faster, you could eliminate >75% of the runtime by preserving the pickle cache somewhere, perhaps in the default, sphinx-configured spot. I do agree though that icky statefulness may not be worth the speedup.

Runtimes in that case: (manual runs without the -E option)

 # | pseudoxml | dummy |
--- | -------------- | ---------- |
1   | .572          | .493      |
2   | .568          | .490 *    |
3   | .545          | .485      |
4   | .561          | .477      |
5   | .563 *        | .495      |

The difference between the two is now 13%, but the difference to the uncached case is 86%.

In summary, I don't think the writing is a big enough issue to reject my branch. It's a 9% slowdown, and isn't the deciding factor for the rmrf. It's Good Enough and correct, although it could be a bit faster with some upstream work.

lcd047:
> To correct this, change your stub of `.write` to a stub of `.write_doc`:

Done, thank you.  Commit a48916c.

> If you actually wanted to make it faster, you could eliminate >75% of the runtime by preserving the pickle cache somewhere, perhaps in the default, sphinx-configured spot.

That's the point of [this](https://github.com/scrooloose/syntastic/blob/a48916ce60b19d51eb0fc7fccb202f2d457c4253/syntax_checkers/rst/sphinx.vim#L23-L28): the cache is preserved until Vim is restarted.  Of course, that's not the same as "full" persistence, but it should speed things up as long as you don't quit Vim.

> Runtimes in that case: (manual runs without the -E option)

That's the point of moving `-E` to [`args`](https://github.com/scrooloose/syntastic/blob/a48916ce60b19d51eb0fc7fccb202f2d457c4253/syntax_checkers/rst/sphinx.vim#L48): you can set `g:syntastic_rst_sphinx_args` to `-n`, and have the same thing in Vim.

> In summary, I don't think the writing is a big enough issue to reject my branch. It's a 9% slowdown, and isn't the deciding factor for the rmrf. It's Good Enough and correct, although it could be a bit faster with some upstream work.

Ok, I'll merge your patch.  Out of curiosity, how does the above results change if you set `g:syntastic_rst_sphinx_args` to `-n` and you do the measurements without quitting Vim, as I pointed out above?

bukzor:
What am I meant to do when I see "syntastic: error: checker rst/sphinx returned abnormal status 1"?

Currently I edit the plugin to echo the command so I can try it myself, but that doesn't seem like it could be recommended practice. After some debugging I found that I had set the sphinx_args to "n" >.<

I'm uncertain how I'd time the command within VimL, but there is a definite speedup after the first check which seems commensurate with the above measurements. I made the above tables by running the sphinx-build command by hand.

lcd047:
Ok, I merged your version, with a few further cleanup changes.  Commit 131ec84.

> What am I meant to do when I see "syntastic: error: checker rst/sphinx returned abnormal status 1"?

Sadly, I don't have a good answer to that.  You can enable debugging: set `g:syntastic_debug` to 3 (the useful values for this particular checker are 1, 3, and 33), then run the checker, and run `:mes` to see what's going on.  But this is not exactly user-friendly.

The problem here is that syntastic doesn't know in advance if the output from the checker is valid or not, and by the time it finds out it's too late to re-parse it to give a better meaningful message.  So, things like a messed up option or a missing extension lead to the message you're seeing.  Yes, there should be a (much) better way to address that than digging your way through the logs.  I don't have a solution to it yet. _shrug_

> I'm uncertain how I'd time the command within VimL

The debug logs will give you some basic timing information, but you could also enable [profiling](https://github.com/scrooloose/syntastic/wiki/Profiling-syntastic).

PR #1456, opened 2015-06-25
Path substitution for g:syntastic_java_javac_custom_classpath_command

ZDBioHazard:
This patch allows `%FILE_PATH%`, `%FILE_NAME%` and `%FILE_DIR%` to be used in `g:syntastic_java_javac_custom_classpath_command` to refer to the current file's full path, basename, and directory, respectively.

This came about because I had written a helper script to build my Android SDK classpath by inspecting the build files in the project directory for the API target. Problem was if I didn't `cd` into the project directory before starting Vim, or if I had files from multiple projects open, the script wouldn't know which project to inspect because it couldn't know what file it was building a classpath for.

### DIFF
--- a/syntax_checkers/java/javac.vim
+++ b/syntax_checkers/java/javac.vim
@@ -129,7 +129,14 @@ function! SyntaxCheckers_java_javac_GetLocList() dict " {{{1
 
     " load custom classpath {{{2
     if g:syntastic_java_javac_custom_classpath_command !=# ''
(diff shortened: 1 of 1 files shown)


lcd047:
@troydm Any comments on this?

troydm:
@lcd047 useful thing, merge it, I never had this problem because when I work with project I always cd into the folder but it seems most people don't do that so yeah this kind of functionality is needed

lcd047:
Thank you.  Merged, with this patch on top: c39563d.

Issue #1455, opened 2015-06-24
Syntastic breaks compilation of Ada code


ampoffcom:
When opening a adb-source file with vim 7.4 and syntastic enabled, the next compilation fails with a linker error.

How to reproduce

a) Syntastic disabled
1. Create a new or open an existing adb file with vim
2. write without changing something
3. quit
4. gnatmake test.adb
5. success

b) Syntastic enabled
1. Create a new or open an existing adb file with vim
2. write without changing something
3. quit
4. gnatmake test.adb
5. linker error with "undefined reference to `main'"

When editing a file that contains errors, e.g. missing semi-colon, syntastic displays an error message. This is expected behaviour.

lcd047:
There is nothing in syntastic that changes the contents of the files being checked.  If you're seeing changes, this is either the result of an interaction with another plugin, some unintended interaction with Vim's UI, or a bug in Vim.  Can you reproduce the problem if you disable __all__ other plugins except syntastic?  Can you please post a minimal file that illustrates the problem?

ampoffcom:
Update: I noticed when I :w the file, vim tries to compile the file, i.e. there is a .ali and a .o-file, but no compilation. When I delete both files gnatmake succeeds.

All plugins disabled except vundle, see vimrc here https://gist.github.com/ampoffcom/ed5ad53e5422cb54eb37

Minimal example here https://gist.github.com/ampoffcom/938df72119a72ad617bf

lcd047:
By default, syntastic runs something like this against your files:
```sh
gcc -c -x ada -fsyntax-only file.adb
```
I would have (naively) assumed that `-fsyntax-only` tells `gcc` not to produce any output files, but apparently I was wrong, this indeed produces `.o` and `.ali` files.  I'm afraid I don't know anything about Ada, or the Ada backend of gcc.  If there is a way to disable the generation of these files, I'd like to know about it, to add it to syntastic.  Perhaps ask on some Ada forum?  Not sure what else to tell you; sorry I'm not of much help.

ampoffcom:
Thanks for your reply. I'll investigate further.

lcd047:
Fixed in dbb28cd.

PR #1457, opened 2015-06-25
Bugfix: fixed Ada syntax checking

ampoffcom:
Related issue: https://github.com/scrooloose/syntastic/issues/1455


lcd047:
Thank you.  Fixed in dbb28cd.

PR #1459, opened 2015-06-25
Security fix in java custom classpath command handler.

ZDBioHazard:
Aww man, moments after I posted PR #1456 I remembered you need to use `shexpand()` when using `system()`. I pushed a fixed patch really fast, but I guess I was too slow.

Anyway, here's a fix for it. Then people can name their files silly things like `foo-$(touch something.txt).java` without issues.


lcd047:
I was wondering about that.  Fixed in dc15fa0.

PR #1463, opened 2015-06-29
Erlang: Translate relative paths from non-root rebar.config files

lavrin:
The checker couldn't find includes when editing files in a repository with a release-like layout (in contrast to an app-like layout where everything was fine).
To flesh the problem out, let's take the following layout for example:

```
.
├── apps
│   ├── ejabberd
│   │   ├── include
│   │   ├── rebar.config
│   │   └── src
│   └── ejabberd_extensions
│       ├── rebar.config
│       └── src
├── rebar
└── rebar.config
```

When Vim's CWD is set to the project root (`.`) and a file under `apps/ejabberd_extensions/src` is edited, the syntax checker can't properly compile the file if it includes headers from `apps/ejabberd/include`, even though `apps/ejabberd_extensions/rebar.config` contains a proper `{i, "../ejabberd/include"}` flag for `erlc`.
This stems from the path in `rebar.config` being relative to the file location (which **is not** at project root), whereas `compile:file/2` interpreting it as relative to CWD (which **is** the project root).
The approach taken in this PR is to translate any relative path found in `{i, _}` option to equivalent absolute path using the location of `rebar.config` the option is found in.

Whew, I hope you get the point - it didn't work for me, but it does now ;) Some people who drop by here, use Erlang and could verify the PR are probably @robashton, @russss or @ahmadsherif.

### DIFF
--- a/syntax_checkers/erlang/erlang_check_file.erl
+++ b/syntax_checkers/erlang/erlang_check_file.erl
@@ -10,10 +10,17 @@ main([File]) ->
             warn_unused_import,
             report,
             {i, Dir ++ "/include"}],
+    %% `rebar.config` is looked for,
+    %% but it is not necessarily the one in the project root.
+    %% I.e. it may be one deeper in the project file hierarchy.
     RebarFile = rebar_file(Dir),
+    %% `rebar.config` might contain relative paths.
+    %% They are relative to the file! Not to the project root.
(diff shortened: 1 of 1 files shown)


lcd047:
Thank you.  @robashton, @russss, @ahmadsherif - Any comments on this?

ahmadsherif:
Looks fine to me.

lcd047:
Ok, merged, thank you.

PR #1474, opened 2015-07-15
Fix #1473

### LINKED ISSUE #1473: clang_check and clang_tidy checkers do not work automatically with compilation databases
The current implementation assumes that no compilation database is used since it explicitly adds `--`.

While a user could overwrite `post_args` with an empty string to make the plugin work with a compilation database, the settings for `clang_check` would not disable color-diagnostics.

Suggested fix for clang checkers: Move current `post_args` of `clang_check` and `clang_tidy` to `args`, but allow overwriting the `post_args` with a user-set string. For users not using compilation databases a configured `post_arg` of at least `--` would enable the plugins.

Suggested fix for `oclint`: Move to `args` and drop `--`.

bbannier:
This should fix #1473.

### DIFF
--- a/syntax_checkers/c/clang_check.vim
+++ b/syntax_checkers/c/clang_check.vim
@@ -27,14 +27,14 @@ set cpo&vim
 
 function! SyntaxCheckers_c_clang_check_GetLocList() dict
     let makeprg = self.makeprgBuild({
+        \ 'args':
+        \   '-extra-arg -fshow-column ' .
+        \   '-extra-arg -fshow-source-location ' .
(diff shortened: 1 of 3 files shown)


lcd047:
These don't seem to be needed: setting `g:syntastic_c_clang_check_post_args`, `g:syntastic_c_clang_tidy_post_args`, and `g:syntastic_c_oclint_post_args` to empty strings (and the same for C++) seem to enable `clang_check`, `clang_tidy`, and `oclint` to use compilation databases.  Sorry, but no.

PR #1480, opened 2015-07-19
Add ruby flog checker

pixelastic:
Hello,

I've added a syntax checker for ruby using [flog](https://github.com/seattlerb/flog). Flog gives each method a score. The higher the score, the harder to maintain/understand/debug the method is. Flog is used, for example, by [Code Climate](https://codeclimate.com/) in calculating their overall complexity score.

The syntax checker allows for two configuration options (`g:syntastic_ruby_flog_threshold_warning` and `g:syntastic_ruby_flog_threshold_error`). If the first threshold is reached, a warning will be generated. If the second is reached, this will be an error. The default (arbitrary) values are `45` and `95`.

### Limitations

Flog itself is confused when the code uses the `%w()` notation for array of words. It then output weird line numbers. This syntax checker does not solve that issue so you might have errors displayed on the wrong line.

Flog itself is slow. It takes ~0.4s on my machine to run (`$ time flog file.rb`), so it creates a noticeable delay when activated on save.

### DIFF
--- a/syntax_checkers/ruby/flog.vim
+++ b/syntax_checkers/ruby/flog.vim
@@ -1,0 +1,82 @@
+"============================================================================
+"File:        flog.vim
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Tim Carry <tim at pixelastic dot com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"
+"============================================================================
+
+if exists('g:loaded_syntastic_ruby_flog_checker')
+    finish
+endif
+let g:loaded_syntastic_ruby_flog_checker = 1
+
+if !exists('g:syntastic_ruby_flog_threshold_warning')
+    let g:syntastic_ruby_flog_threshold_warning = 45
+endif
+
+if !exists('g:syntastic_ruby_flog_threshold_error')
+    let g:syntastic_ruby_flog_threshold_error = 90
+endif
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_ruby_flog_GetLocList() dict
+    let makeprg = self.makeprgBuild({})
+
+    " Example output:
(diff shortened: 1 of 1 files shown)


lcd047:
Thank you.  This is one of the most unpleasant checkers I had to deal with, mainly because of the bizarre choice of the authors to report complexity as a floating point number, and the crappy support for said numbers in Vim.  Merged, with this patch on top: 6929f24.

pixelastic:
Who, thanks for going the extra mile and the voodoo work on the float casting.

Could you share some light on the `'%\m%\s%#%m: %.%# %f:%l'` `errorformat`? What are `%\m` and `%\s`?

lcd047:
`%\` is `\`, and the rest follows from that: `%\m` is `\m`, "magic" regexps, and `%\s%#` is `\s*`, zero or more spaces.  `%f` is the filename, and `%.%#` skips the function names.  Your `%p` was wrong, it's supposed to match things like `___^` and infer a column number from that.

pixelastic:
Thanks. I had trouble understanding how to use regular regexp patterns in the `errorformat`, this helps.

lcd047:
I suppose I should write a wiki page about this at some point.  Here's a (slightly modified) excerpt from a private exchange with a Vim developer:

> The idea behind `errorformat` is pretty simple: it's a list of "augmented" regexps, which are used to parse compiler errors and transform them into `qflist`s.  Each augmented regexp may contain a number of `%`-formats, which are transformed into regexp groups `\(...\)` when `errorformat` is parsed (so the result of parsing `errorformat` is actually a list of plain regexps).  When one of these regexps is matched against a compiler error line, the values matched by the `\(...\)` groups are written to the fields of a `qflist` item, each field corresponding to an `%`-format.  And that's basically all there is to it, the rest are flags for ignoring patterns and joining lines, and for changing the order of matching.  I wrote a [short blurb](http://stackoverflow.com/a/29102995/1658042) at StackOverflow that clarifies a few other points.

pixelastic:
I read your post on StackOverflow already, to better understand how this was working. I also had a look at `errorformat` in other syntax checkers but the multiline syntax was adding more confusion.

I guess a few patterns and examples should clarify things. Maybe simple examples using `%l` and `%f`, and going gradually to more complex ones using custom regexp, ignoring parts, going multiline, etc.

I would write the wiki page myself if I was more seasoned in `errorformat` handling.

Issue #1485, opened 2015-07-23
hdevtools call uses full path - fails to recognize local modules


sdx23:
The Haskell hdevtools-checker is called using the full path of the file. This leads to a false positive report, that local modules (Module.hs in the same directory as the source file to be checked) are unknown ("Could not find module ‘Module’ Use -v to see a list of the files searched for.").

Removing the fname-declaration from the makeprg fixes this, however I'm not sure why it's there in the first place.

vim: 7.4.712
syntastic: current git HEAD
hdevtools: version 0.1.0.9 (ghc-7.10.1-x86_64-linux, cabal-1.22.2.0)

lcd047:
> I'm not sure why it's there in the first place.

Ah, good thing you asked before writing a patch.  Wait, never mind. ;)

The filename is there because you can in principle open the file from some other directory, and `hdevtools` should still do something sensible.  If you don't want the filename you can override that by setting:
```vim
let g:syntastic_haskell_hdevtools_fname = ''
```
You don't need to patch the sources.

sdx23:
Ah, I see. However, even in that case hdevtools wouldn't work. An actual fix'd be to run it with working-directory being that of the source file.

sdx23:
Actually, after checking that now, I have to disagree: hdevtools works fine with 

let g:syntastic_haskell_hdevtools_fname = ''

even when opening files in a relative path. The only problem are absolute paths (then the same issue as in the beginning emerges).

lcd047:
> An actual fix'd be to run it with working-directory being that of the source file.

Then error messages referring to other files would point to the big void... :)

PR #1486, opened 2015-07-23
fix hdevtools fname to enable recognizing local modules (issue #1485)

sdx23:
Fix for issue #1485


lcd047:
This is not needed.  The same effect can be achieved with:
```vim
let g:syntastic_haskell_hdevtools_fname = ''
```

PR #1491, opened 2015-07-28
New checker for qml: qmllint

Lekensteyn:
Note that some syntax errors generate a line number past the last line:

    file.qml:2 : Expected token `}'

(Example contents: `Item {`)

Then unfortunately no marker is displayed on the side. I observed this with other file types as well. Maybe GetLocList should check whether a generated line number equals `n + 1` for `n` line numbers, and in that case set the line number to `n` instead (but only if no error exists at that line). Hmm.


lcd047:
Thank you.  Merged, with this patch on top: 9d69f4c.  Please add some references to the corresponding [wiki page](https://github.com/scrooloose/syntastic/wiki/QML%3A---qmllint).

Lekensteyn:
Thanks, does the edited wiki page look good to you?

lcd047:
Thank you.  I rephrase it a little, to be more in line with the other pages.

PR #1509, opened 2015-08-07
lslint: add checker for lsl

tecywiz121:
Adding a syntax checker for the [Linden Scripting Language][0], used in Second Life and other similar games.

You can get lslint [here][1].

[0]: https://en.wikipedia.org/wiki/Linden_Scripting_Language
[1]: https://github.com/Makopo/lslint


lcd047:
This seems a little too specialized to be added to syntastic.  Perhaps you can maintain it as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external)?

PR #1512, opened 2015-08-10
Suppresses Rubocop warnings to get accurate version

buccolo:
Hello,

First of all, thank you so much for this awesome tool! :beers: 

I've noticed my Rubocop install outputs this warning:
```bash
$ rubocop --version
warning: parser/current is loading parser/ruby21, which recognizes
warning: 2.1.7-compliant syntax, but you are running 2.1.6.
0.32.0
```

Which messes up the `getVersion` function.

Because there's [no flag in Rubocop to hide such warnings](https://github.com/bbatsov/rubocop/issues/1819), I've ignored them like it's done in [coffee.vim](https://github.com/scrooloose/syntastic/blob/a7fde99ea9a4eb26e6dfedb3ea54bbeefb71b325/syntax_checkers/coffee/coffee.vim#L28-L29)

Please let me know if there's anything else I can/should do.

Thanks!

### DIFF
--- a/syntax_checkers/ruby/rubocop.vim
+++ b/syntax_checkers/ruby/rubocop.vim
@@ -22,7 +22,8 @@ function! SyntaxCheckers_ruby_rubocop_IsAvailable() dict
     if !executable(self.getExec())
         return 0
     endif
-    return syntastic#util#versionIsAtLeast(self.getVersion(), [0, 12, 0])
+    let ver = self.getVersion(self.getExecEscaped() . ' --version 2>' . syntastic#util#DevNull())
+    return syntastic#util#versionIsAtLeast(ver, [0, 12, 0])
 endfunction
 
 function! SyntaxCheckers_ruby_rubocop_GetLocList() dict


lcd047:
There's always the radical approach of reading the docs. :) In this case, the relevant [wiki page](https://github.com/scrooloose/syntastic/wiki/Ruby%3A---rubocop).

buccolo:
Hmmm thanks @lcd047 I missed that :+1: 

I think this PR is still valid, because we suppress `rubocop` warnings only to get the current version.

I shouldn't wrap my `rubocop` in a script that silences all warnings, because the wiki page mentions there are useful warnings:

> Most issues returned by rubocop are warnings, so for best results your g:syntastic_quiet_messages should not filter out warnings.

Additionally, some folks (including me) can't update Ruby to the latest version as fast as we would like (external requirements, etc..). So we shouldn't use `rbenv` (or similar tools) to use the latest version of Ruby, as the wiki suggests, because it would break our [dev/prod parity](http://12factor.net/dev-prod-parity) a little bit.

What do you think?

lcd047:
> I think this PR is still valid, because we suppress `rubocop` warnings only to get the current version.

No it isn't.  It also ignores warnings unrelated to Ruby version.

> I shouldn't wrap my `rubocop` in a script that silences all warnings,

The [wrapper](https://gist.github.com/lcd047/96138909015f2f8d2d36) I suggest doesn't ignore _all_ warnings.

> because the wiki page mentions there are useful warnings:

You're mixing warnings about the abnormal functioning of rubocop with warnings about the contents of the file being checked.  The former are fatal as far as syntastic is concerned.  The latter are normal output.

> Additionally, some folks (including me) can't update Ruby to the latest version as fast as we would like (external requirements, etc..). So we shouldn't use `rbenv` (or similar tools) to use the latest version of Ruby

If you have several versions of Ruby installed, you can use `rbenv` or `rvm` to switch between them without making a mess.  This is orthogonal to whether the versions you are using are old or new.

> What do you think?

I think you should take your time to make sense of all this.

PR #1517, opened 2015-08-13
fix jshint and jsxhint loading error

carlosliu:
In OSX and Linux, jshint and jsxhint put their version string output in stderr instead of stdout.
Therefore, ```system('jshint --version')``` return empty value, which ends up being a version parse error.
```syntastic: error: checker javascript/jshint: can't parse version string (abnormal termination?)```
Of course, Windows environment don't have this issue.

### DIFF
--- a/syntax_checkers/html/jshint.vim
+++ b/syntax_checkers/html/jshint.vim
@@ -23,7 +23,8 @@ function! SyntaxCheckers_html_jshint_IsAvailable() dict
         return 0
     endif
 
-    let ver = self.getVersion(self.getExecEscaped() . ' --version 2>&1')
+    let command = self.getExecEscaped() . ' --version'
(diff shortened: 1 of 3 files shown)


lcd047:
> Therefore, `system('jshint --version')` return empty value

Have you actually tried this with a working `jshint`?  Unless you have messed with the (many) settings affecting `system()`, Vim is supposed to do the redirection automatically.

On a side note: please don't bother reporting any problems if you changed any of these from the defaults: `shellcmdflag`, `shellpipe`, `shellquote`, `shellredir`, `shellxquote`, `shellxescape`.

> which ends up being a version parse error.

Something else is going on, and it has to do with your `jshint`, not with syntastic.  Upgrade to the latest master HEAD, open a JavaScriot file, set `g:syntastic_debug` to 33, run `:SyntasticInfo`, and post the output.

carlosliu:
Hey, sorry, my fault. It is a shell issue.

PR #1518, opened 2015-08-13
Remove obselete tslint filename flag

nikklassen:


lcd047:
The only thing this patch accomplishes is break compatibility with older versions of `tslint`. Sorry, but no.

nikklassen:
Sorry about that oversight.  I should've thought through my change more before I made the PR.  I'd still like to see this fixed, so I'd be happy to add a check for the tslint version to decide whether or not to include the flag.  Can we reopen this PR or would you like me to make another once I fix that issue?

lcd047:
It has been fixed in 748725c.  You patched a version that was already fixed.

PR #1522, opened 2015-08-19
Add support for JSHint within HTMLDjango files

rummik:
This PR copies `syntax_checkers/html/jshint.vim` to `syntax_checkers/htmldjango/jshint.vim`, and replaces references to `html` with `htmldjango`


lcd047:
This is the wrong approach for at least two reasons.  First, because syntastic can handle composite filetypes (see `:help syntastic-composite`).  Thus if you set the filetype for your file to `html.django`, you can check it with `jshint` directly.  And second, because you should [never copy checkers](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#duplicates) like this.  Sorry, but no.

PR #1524, opened 2015-08-24
Add go errcheck syntax checker

EllisV:
Go Errcheck Syntax Checker
========================

Add support for go errchecker syntax check. Errcheck outputs violations in folowing format: 

```
[filename]:[line]:[column][whole line]
```


lcd047:
The problem with running errcheck from syntastic was that errcheck only produced meaningful results when run from the relevant directory.  Did that change when I wasn't looking?

lcd047:
I take it nothing changed?  Well then, perhaps this is better left out of syntastic.

PR #1526, opened 2015-08-26
Java checkstyle: Add support for property file

ingemaradahl:
More elaborate checkstyle configurations sometime depend on the property file being set. The added stuff is optional, so unless you specify the new variable, nothing will change.

If the PR is accepted, we shouldn't forget to update https://github.com/scrooloose/syntastic/wiki/Java%3A---checkstyle

### DIFF
--- a/syntax_checkers/java/checkstyle.vim
+++ b/syntax_checkers/java/checkstyle.vim
@@ -49,12 +49,19 @@ function! SyntaxCheckers_java_checkstyle_GetLocList() dict
         let fname = substitute(syntastic#util#system('cygpath -m ' . fname), '\m\%x00', '', 'g')
     endif
 
+    let checkstyle_args = [
+        \ '-c', expand(g:syntastic_java_checkstyle_conf_file, 1),
+        \ '-f', 'xml']
+
+    if exists("g:syntastic_java_checkstyle_property_file")
+        let property_file = expand(g:syntastic_java_checkstyle_property_file, 1)
+        let checkstyle_args += ['-p', property_file]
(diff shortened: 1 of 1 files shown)


lcd047:
You can achieve the same thing without a patch:
```vim
let g:syntastic_java_checkstyle_args_after = ['-p', '/path/to/file']
```
Thus, sorry but no.

ingemaradahl:
Nope, that doesn't work. I'm not sure if the way the checkstyle checker sets up it's args are up to scratch, but judging by how it calls makeprgBuild as
```vim
        \ 'args_after': [
        \       '-cp', expand(g:syntastic_java_checkstyle_classpath, 1),
        \       'com.puppycrawl.tools.checkstyle.Main'] ...
```
it may be the case that setting `g:syntastic_java_checkstyle_args_after` doesn't make a difference, as it's overwritten anyway? I.e, is it the case that setting the `args_after` entry should take in to consideration `g:syntastic_java_checkstyle_args_after` variable, if it is set?

lcd047:
You can see exactly what syntastic is doing by turning on debugging.  Set `g:syntastic_debug` to 1, run the checker, and run `:mes`.

ingemaradahl:
Yes, I did that. The stuff I put in `g:syntastic...args_after` isn't in the command line.

lcd047:
Sorry, it's called `g:syntastic_java_checkstyle_post_args`, not `*_args_after`.

ingemaradahl:
That did the trick! Thanks for taking the time to help me out.

PR #1527, opened 2015-08-27
Set the GOGC env var to off to speed up go builds

gsalgado:
With GO 1.5, builds are up to 2x slower, so this becomes really important. It
was suggested on https://twitter.com/mitchellh/status/634419559252914176


lcd047:
Merged, thank you.

PR #1531, opened 2015-08-31
Added Purescript language and Pulp syntax checker

seanhess:
I've tested with as many different kinds of errors I can generate, and it appears to be useful. Please let me know if there's anything you'd like me to change.

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -65,6 +65,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'po':            ['msgfmt'],
         \ 'pod':           ['podchecker'],
         \ 'puppet':        ['puppet', 'puppetlint'],
+        \ 'purescript':    ['pulp'],
         \ 'python':        ['python', 'flake8', 'pylint'],
         \ 'qml':           ['qmllint'],
         \ 'r':             [],
--- a/syntax_checkers/purescript/pulp.vim
+++ b/syntax_checkers/purescript/pulp.vim
@@ -1,0 +1,48 @@
+"============================================================================
+"File:        pulp.vim
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Sean Hess <seanhess at gmail dot com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"
+"============================================================================
+
+if exists('g:loaded_syntastic_purescript_pulp_checker')
+    finish
+endif
+let g:loaded_syntastic_purescript_pulp_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_purescript_pulp_GetLocList() dict
+
+    let makeprg = self.makeprgBuild({
+        \ 'exe_after': 'build --main ',
+        \ 'fname': syntastic#util#shexpand('%:p') })
+
+    let errorformat =
+        \ '%E\\s%#\"%f\"%.%#,' .
+        \ '%E%.%#Error at %f line %l\, column %c - %.%#,'.
+        \ '%ZSee http%.%#,' .
+        \ '%C\\s%#%m,' .
+        \ '%Z'
+
+
+    return SyntasticMake({
+        \ 'makeprg': makeprg,
+        \ 'errorformat': errorformat })
+
+endfunction
+
+call g:SyntasticRegistry.CreateAndRegisterChecker({
+    \ 'filetype': 'purescript',
+    \ 'name': 'pulp'})
+
+let &cpo = s:save_cpo
+unlet s:save_cpo
+
+" vim: set sw=4 sts=4 et fdm=marker:


seanhess:
This has been tested by other community members now too.

sharkdp:
This is great! :+1: 

 I believe it does need some work, though:

* Using `pulp build --main ..` will not work in all cases. When editing a file in the `test` directory of a pulp project, `pulp build` will only compile the files in `src` and ignore the `test` directory. What about non-pulp projects?
* A standard type mismatch error gets displayed as `Cannot unify typePrim.Intwith typePrim.String` (missing whitespace)
* There are a few errors which are not recognized / rendered properly: e.g.: [Orphan instances errors](https://github.com/purescript/purescript/blob/5fdc97791ae74794e80692fedcfdf8952dfe2f7f/src/Language/PureScript/Errors.hs#L612-L613)
* This is really more a problem of syntastic, but a full purescript build (via pulp) can be quite long (1s and more). Since pulp is run synchronously and blocks vim, this makes it almost unusable as a syntax checker.

lcd047:
Aside from the issues pointed out by @sharkdp, `errorformat` is wrong.  Without looking at what this is supposed to match, the following fixes the syntax errors:
```vim
    let errorformat =
        \ '%E%\s%#psc: %m,' .
        \ '%C%\s%#at "%f" (line %l\, column %c)%.%#,' .
        \ '%E%\s%#"%f" (line %l\, column %c)%.%#,' .
        \ '%E%.%#Error at %f line %l\, column %c - %.%#,'.
        \ '%-G%\%# ERROR: Subcommand%.%#,' .
        \ '%\%# ERROR: %m,' .
        \ '%-G%\s%#See http%.%#,' .
        \ '%C%\s%#%m'
```
I'd suggest the following framework for testing `errorformat`s:
```vim
let &errorformat = '...'
lgetexpr [
    \ 'psc: No files found using pattern: src/**/*.js',
    \ 'Error found:',
    \ 'Error at /home/lcd047/tmp/purs/src/Main.purs line 5, column 4 - line 5, column 4:',
    \ '  Unable to parse module:',
    \ '    unexpected ]',
    \ '    expecting binder, | or ='
    \ 'See https://github.com/purescript/purescript/wiki/Error-Code-ErrorParsingModule for more information, or to contribute content related to this error.',
    \ ]
echomsg string(getloclist(0))
lopen
```

seanhess:
@sharkdp: I'm new to Purescript, `pulp build` is how I learned how to do it. The syntax checker is called `pulp` on purpose. If you don't have pulp installed it won't do anything. Would it be faster to use `psc` directly? What would the correct thing to type on the command-line to compile a single file and have all of its dependencies available?

> A standard type mismatch error gets displayed as Cannot unify typePrim.Intwith typePrim.String (missing whitespace)

There's no way to display multiple lines in syntastic errors that I am aware of. @lcd047 is that correct? Is there another way you'd like to see that error on one line? It seems readable to me. 

How can I generate an orphan instance?

seanhess:
@lcd047 That's super useful thanks! It's such a pain to test the errorformat on a real project. 

You are getting syntax errors in the current errorformat? How can I see them? I definitely generated some earlier while developing it, but I don't see any now and it appears to work. I'm happy to fix them but it would be difficult without being able to reproduce them.

lcd047:
> There's no way to display multiple lines in syntastic errors that I am aware of. @lcd047 is that correct?

It depends on what you mean by that.  Loclist windows don't display multiline text fields as multiple lines.  That's a limitation of Vim, not syntastic.  But you _can_ join messages spread across multiple lines, and format them to look reasonable.

> You are getting syntax errors in the current errorformat?

Like I said, I haven't run that in Vim.  But some of constructs you're using don't work the way you seem to expect.

Anyway, looking at the list of possible errors, I don't think you stand any chance to cover them all with a reasonable effort.  Not with `errorformat` anyway. _shrug_

sharkdp:
> I'm new to Purescript, pulp build is how I learned how to do it. The syntax checker is called pulp on purpose. If you don't have pulp installed it won't do anything.

Doing nothing - if pulp is not available - is fine. In general though, using pulp as a syntax checker does feel wrong to me:
* Pulp is a build tool and running `pulp build` has side effects (files will be written to the `output` folder). While this might be okay for most cases, a non-suspecting user might be surprised that vim suddenly builds his project if he saves a source file (or creates an `output` folder).
* Pulp relies on a certain project structure (`bower.json` file in the project root)
* As mentioned above, using `pulp build --main` does not work in the `test` directory of a project, because pulp simply builds the project from the `src` folder.

> Would it be faster to use psc directly?

Probably not, since pulp is just a wrapper around `psc`. On the other hand, there might be a chance to speed up the build with something like `psc --no-opts` (disable optimizations, we are not interested in the generated code). I have no experience with this. Optimally, there would be something like `--syntax-check-only` for psc, without any side effects. I'm not sure if something like this exists (/cc @paf31).

> What would the correct thing to type on the command-line to compile a single file and have all of its dependencies available?

It's actually not straightforward because we have to include all of the projects purescript and javascript files. This is where using pulp comes in handy.

Alternatively, I was thinking it would be really nice to have something like [hlint](https://github.com/ndmitchell/hlint) for purescript. This could potentially be much faster than a full build of the project (?). I realize that I'm just piling up ideas here... that's not very helpful, sorry.

> > A standard type mismatch error gets displayed as Cannot unify typePrim.Intwith typePrim.String (missing whitespace)

> There's no way to display multiple lines in syntastic errors that I am aware of. @lcd047 is that correct? Is there another way you'd like to see that error on one line? It seems readable to me.

If it's possible to concat these lines with some whitespace in between, that would be great.

> How can I generate an orphan instance?

There is an example on the purescript wiki [here](https://github.com/purescript/purescript/wiki/Error-Code-OrphanInstance). When using all of this code in a single file, your syntastic checker actually works nicely (correctly indicating the orphan instance error). So ignore my comment above, this is not due to the error formating.

However, when I split this code into separate modules `A.purs`, `B.purs` and `C.purs`, the checker incorrectly shows a syntax error in module `A.purs`  on line 7 (there is no line 7 in `A.purs`). So it's probably a matter of parsing the output for syntax errors in the current file.

paf31:
There is no way to run the compiler without generating output files right now, at least not using the standard executables. An option could be added easily though and I'd happily take a PR to add a `--check-only` option or something like that. Would we want to run the typechecker or just check things like imports? Skipping type checking or separating it into its own build command would speed things up quite a lot, if that's possible.

Since the compiler accepts globs now, it might be easier to use `psc` directly, and then things like whether to include the `test` directory become simply a matter of changing a glob in the configuration.

lcd047:
@paf31 As far as I'm concerned building the project while checking is annoying, but not fatal.  Other checkers do that, and people still use them.  The real show stopper however is having to parse error messages from 100+ ad-hoc formats.  It would be nice if psc would have an option to output errors in a machine-friendly format.  I do understand that adding such a feature at this point might take some work, but without it it isn't reasonable to write a syntax checker for an editor. _shrug_

paf31:
That's been requested before. I can prioritize it now that proper editor support is becoming a thing. Is there a best format I can use?

lcd047:
@paf31 Something like `/path/to/file:6.3-6.7: Expression 33 does not have type Prim.String` would be perfect.  JSON would be fine too, and probably more flexible.

One more thing: column numbers should be counted as byte offset (1 tab = 1 character).  Otherwise, the exact position of the error depends of tab size...

seanhess:
Do we want to try to push for something we can merge short-term, or wait until you make changes to psc @paf31? We're probably going to end up using Elm for our project so I would prefer to get this to a useful state, get it merged, and let you guys make PRs to it as you improve things.

lcd047:
@seanhess The current code has various problems, and the changes kindly offered by @paf31 would make a big difference.  If you need a checker right now, perhaps you can maintain it [externally](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external) until the relevant features are added to `psc`?

seanhess:
@lcd047 It's already useful. While there are many diverse errors the vast majority fall within the same formats. I would rather clean up the things you mentioned (the errorformat issues) and get this merged while we wait so people can use it. 

But I'm ok either way. I would like to do whatever is more helpful to the Purescript community.

lcd047:
@seanhess Sadly, useful to you != problem-free for me.  Missing errors is a great source of bug reports, and a legitimate one at that. _shrug_

seanhess:
I totally understand. Well I'll defer to you of course. If you want to wait
that's fine. Thanks.

PR #1532, opened 2015-09-01
gotype - parse package test files if checking a test file

jonasi:
gotype wasn't parsing `_test.go` files, which is fine most of the time - except when you were writing test files :)

This pr passes the flag `-a` to `gotype` if we're currently on a `_test.go` file. 

From the `gotype` usage:

```
  -a=false: use all (incl. _test.go) files when processing a directory
```


lcd047:
Thank you.  This should be configurable; commit d077102.

PR #1536, opened 2015-09-02
Add jade-lint

benjaminparnell:
This adds [jade-lint](https://github.com/benedfit/jade-lint), a syntax checker for [jade](http://jade-lang.com/) templates.


lcd047:
Thank you.  Merged, with this patch on top: 703287c.

On a side note: column numbers seem to be produced only when `requireSpacesInsideAttributeBrackets` is `true`, and they are wrong (the numbers are relative to the end of the preceding token, which is useless).

balaclark:
The column number issue is a known bug in the jade-lexer: https://github.com/benedfit/jade-lint/issues/17

PR #1540, opened 2015-09-08
Allow reloading of plugin/syntastic.vim

blueyed:
Use `unlockvar` to allow reloading the plugin, e.g. via scriptease's
`:Runtime`.

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -15,10 +15,12 @@ endif
 let g:loaded_syntastic_plugin = 1
 
 if has('reltime')
+    unlockvar! g:_SYNTASTIC_START
     let g:_SYNTASTIC_START = reltime()
     lockvar! g:_SYNTASTIC_START
 endif
 
+unlockvar g:_SYNTASTIC_VERSION
 let g:_SYNTASTIC_VERSION = '3.6.0-155'
 lockvar g:_SYNTASTIC_VERSION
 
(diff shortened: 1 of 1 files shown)


lcd047:
What problem does this solve?

blueyed:
It allows reloading the plugin, e.g. via scriptease's `:Runtime`.

lcd047:
Yes, that's what `:Runtime` is supposed to do.  What problem does reloading syntastic solve?

blueyed:
> What problem does reloading syntastic solve?

It is useful while hacking on it.  Not a common use case, but I think it should be supported anyway?!

lcd047:
There are more occurrences, 34 in all.  Redefining these variables is generally harmless, although something like the following would be better, since it would keep the original values:
```vim
if has('reltime')
    try
        let g:_SYNTASTIC_START = reltime()
    catch /\m^Vim\%((\a\+)\)\=:E741/
    endtry
    lockvar! g:_SYNTASTIC_START
endif
```
However, reloading syntastic like this is a really bad idea.  Because of the way objects are handled, you'll end up with objects defined in the old run mixed with objects defined in the new one.  To make it safe, you might try `unlet`ing the various variables related to syntastic from `g:`, from `b:` for all loaded buffers, from `w:` for all windows, and maybe others.  This is not trivial.  I tried doing something related a while ago (in order to add a feature to issue warnings when you define a variable `g:syntastic_*` that has no meaning to syntastic), and I gave up as not worth the effort.

blueyed:
Ok, I see.

So there's no easy way to make it easily reloadable?

lcd047:
Definitely not easy, no.

blueyed:
Too bad, closing this PR then for now.

PR #1541, opened 2015-09-08
Skip ignored files in BufEnterHook

blueyed:
Without this an existing location list would pop up again, although
`b:syntastic_skip_checks` is set to 1 (after the list has been created
initially).

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -270,6 +270,11 @@ function! s:BufEnterHook() abort " {{{2
     call syntastic#log#debug(g:_SYNTASTIC_DEBUG_AUTOCOMMANDS,
         \ 'autocmd: BufEnter, buffer ' . bufnr('') . ' = ' . string(bufname(str2nr(bufnr('')))) .
         \ ', &buftype = ' . string(&buftype))
+
+    if s:_skip_file()
+        return
+    endif
+
     if &buftype ==# ''
         call s:notifiers.refresh(g:SyntasticLoclist.current())
     elseif &buftype ==# 'quickfix'


lcd047:
What is the exact scenario of the error window opening when it shouldn't?  The patch is wrong, it prevents automatic closing of error windows (`&buftype == 'quickfix'`).

blueyed:
> What is the exact scenario of the error window opening when it shouldn't?

Errors were collected and I've set the buffer to be ignored (`b:syntastic_skip_checks=1`).  Then I closed the location list manually, but when going back to the window, it gets opened again.  I am using `g:syntastic_auto_loc_list = 1`.

Should the skipping only get done for `&bt == ''`?

lcd047:
It would be better to make the refresh function for `autoloclist` keep state.  Currently, if you have `g:syntastic_auto_loc_list` set to 1, you can't close the quickfix window if the quickfix window has focus (try running `:lclose`).  Keeping state would take care of both problems, but doing that is not completely trivial.  I'll try to think of something.

blueyed:
Great, I'd be happy to try what you come up with!

lcd047:
I'm afraid I don't have a good fix for this.  The reasons are rather complicated.  Making the refresh keep state is not possible because of limitations of the Vim API.  Making syntastic not re-open the window after you set `b:syntastic_skip_checks` would involve either changing the `loclist`, or making the refresh function aware of `b:syntastic_skip_checks`.  The former would break the design assumption that the `loclist` doesn't change after being created.  The latter would break the design assumption that state variables don't change between some checkpoints.  Basically the current behaviour is technically correct, even though it isn't what you want.

As a workaround, I suggest you to run `SyntasticReset` instead of `lclose`, then set `b:syntastic_skip_checks`, and run the checker again.  Sorry about that.

PR #1548, opened 2015-09-15
Hlint: handle hint file parsing errors.

supki:
When hints (including those in ANN pragmas in Haskell source) fail to
parse, Hlint uses a different error format. As parse errors are fatal, ignoring
them effectively mutes the checker.


lcd047:
Merged, thank you.

PR #1550, opened 2015-09-16
add gometalinter syntax checker

joshuarubin:
Adds support for gometalinter syntax checker: https://github.com/alecthomas/gometalinter

### DIFF
--- a/syntax_checkers/go/gometalinter.vim
+++ b/syntax_checkers/go/gometalinter.vim
@@ -19,20 +19,29 @@ let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_go_gometalinter_GetLocList() dict
-    let opts = syntastic#util#var('go_gometalinter_args', '')
+    let opts = syntastic#util#var('go_gometalinter_args', '-t')
     let opt_str = (type(opts) != type('') || opts !=# '') ? join(syntastic#util#argsescape(opts)) : opts
     let makeprg = self.getExecEscaped() . ' ' . opt_str
 
     let errorformat =
-        \ '%f:%l:%c:%m,' .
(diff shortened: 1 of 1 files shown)


lcd047:
Thank you.  Merged, with this patch on top: 8bb535b.

joshuarubin:
The changes you made cause this linter to stop working properly. `gometalinter` needs to be called, not with the file, but using the cwd of the file. I copied the trick used in the [`go` syntax checker](https://github.com/scrooloose/syntastic/blob/master/syntax_checkers/go/go.vim) which is why I ended up using `self.makeprgBuild` as well.

```vim
    " The go compiler needs to either be run with an import path as an
    " argument or directly from the package directory. Since figuring out
    " the proper import path is fickle, just cwd to the package.

    let errors = SyntasticMake({
        \ 'makeprg': makeprg,
        \ 'errorformat': errorformat,
        \ 'cwd': expand('%:p:h', 1),
        \ 'env': {'GOGC': 'off'},
        \ 'defaults': {'type': 'e'} })
```

lcd047:
> `gometalinter` needs to be called, not with the file, but using the cwd of the file.

Sadly, that isn't what `cwd` is supposed to do in `SyntasticMake()`.   Please try f2084e9.

> I copied the trick used in the [`go` syntax checker](https://github.com/scrooloose/syntastic/blob/master/syntax_checkers/go/go.vim)

The `go` checker is a particularly oddball case, and certainly not one you want to use as a template for new checkers.  As I said: next time perhaps ask before starting to write code?

joshuarubin:
The latest change works correctly, in fact, the way it calls gometalinter is *more* correct as some of the tools need the explicit path and weren't working correctly in my version. Thanks for the help.

PR #1553, opened 2015-09-18
JavaScript checkers: Add eslint_d checker

mantoni:
The eslint_d checker is essentially an eslint checker on steroids. It launches a daemon process in the background to only read all the node modules once. This brings the execution times down significantly.

Have a look at the README for more on how eslint_d works: https://www.npmjs.com/package/eslint_d

I'm new to vimscript and would be glad to hear your suggestions. I tried to stay close to the existing eslint checker. Thanks!

### DIFF
--- a/syntax_checkers/javascript/eslint_d.vim
+++ b/syntax_checkers/javascript/eslint_d.vim
@@ -1,0 +1,44 @@
+"============================================================================
+"File:        eslint_d.vim
+"Description: Javascript syntax checker - using eslint
+"Maintainer:  Maximilian Antoni
+"License:     MIT <https://github.com/mantoni/eslint_d.js/blob/master/LICENSE>
+"============================================================================
+
+if exists('g:loaded_syntastic_javascript_eslint_d_checker')
+    finish
+endif
+let g:loaded_syntastic_javascript_eslint_d_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_javascript_eslint_d_IsAvailable() dict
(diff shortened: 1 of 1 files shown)


lcd047:
Syntastic runs checkers at most once per file write rather than each time the buffer changes.  As such, using `eslint_d` with syntastic instead of the plain `eslint` seems rather pointless, and unnecessarily restrictive.  However, if you insist on using it, you can write a [wrapper script](https://gist.github.com/lcd047/68e93a4c42442807b78f) and point `g:syntastic_javascript_eslint_exec` to it:
```vim
let g:syntastic_javascript_eslint_exec = '/path/to/eslint_d.sh'
```
The preferred way to deal with duplicate checkers is to use the existing [duplication mechanisms](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#duplicates) and / or to write wrapper scripts.  Actually writing a duplicate should be left as a very last resort.  The reasons are not only related to good hygiene; avoiding duplicates prevents the variations of the checker from going out of synch with each other.

Also, for what it's worth: the first version of your PR actually worked, this one doesn't. :)

Thus, sorry, but no.

mantoni:
Thanks for the pointers. Version 2.1.0 of `eslint_d` now supports all the required options to directly work with the existing syntastic checker. I've updated my documentation accordingly.

lcd047:
Of course, this is even better.  I added a note to that effect to the [wiki](https://github.com/scrooloose/syntastic/wiki/JavaScript%3A---eslint).

mantoni:
Cool, thanks!

PR #1559, opened 2015-09-21
Checkstyle checker: support classpath file

Scuilion:
Add the ability to externalize the checkstyle classpath. Similar to how it is done in javac.

### DIFF
--- a/syntax_checkers/java/checkstyle.vim
+++ b/syntax_checkers/java/checkstyle.vim
@@ -24,6 +24,14 @@ if !exists('g:syntastic_java_checkstyle_conf_file')
     let g:syntastic_java_checkstyle_conf_file = 'sun_checks.xml'
 endif
 
+if !exists('g:syntastic_java_checkstyle_config_file_enabled')
+    let g:syntastic_java_checkstyle_config_file_enabled = 0
+endif
+
+if !exists('g:syntastic_java_checkstyle_config_file')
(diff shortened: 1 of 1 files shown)


lcd047:
Please don't do that.  The `javac` checker is a mess that needs to be simplified, rather than imitated everywhere else.

The syntastic way to deal with this kind of problems is to delegate as much as possible to the OS, by using environment variables and wrapper scripts.  If you set `g:syntastic_java_checkstyle_classpath` to an empty string you can manage classpath externally, by setting the `CLASSPATH` environment variable.

The Vim way to deal with this kind of problems is to use `autocmd`s.  If you set `autocmd`s for your projects you can update `g:syntastic_java_checkstyle_classpath` automatically.  You can source files from `autocmd`s.

The IDE-minded way to deal with this kind of problems is to install a plugin that allows you to have per-project settings.  There are several such plugins around.

The Java way to deal with this kind of problems is to install Eclim and use that instead of syntastic.  There is no way syntastic will ever compete to Eclipse.

Sorry, but no.

wsdjeg:
hi @Scuilion same issue as yours

lcd047:
@wsdjeg Syntastic supports [external checkers](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).  One way to do what you want is to copy the existing checker elsewhere, apply the patch in the PR, and use it as an external checker.  Another way is to have mnore people express interest in this, and persuade me to merge the PR. :)

PR #1561, opened 2015-09-22
Add `xo` checker

arthurvr:
Adds a new JavasScript/JSX checker called `xo`. Repo: https://github.com/sindresorhus/xo

### DIFF
--- a/syntax_checkers/javascript/xo.vim
+++ b/syntax_checkers/javascript/xo.vim
@@ -1,0 +1,48 @@
+"============================================================================
+"File:        xo.vim
+"Description: JavaScript happiness style
+"Maintainer:  Arthur Verschaeve <contact@arthurverschaeve.be>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"============================================================================
+if exists('g:loaded_syntastic_javascript_xo_checker')
+    finish
+endif
+let g:loaded_syntastic_javascript_xo_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_javascript_xo_IsAvailable() dict
+    if !executable(self.getExec())
+        return 0
+    endif
+    return syntastic#util#versionIsAtLeast(self.getVersion(), [0, 1])
+endfunction
+
+function! SyntaxCheckers_javascript_xo_GetLocList() dict
+    let makeprg = self.makeprgBuild({ 'args_before': '--compact'  })
+    let errorformat =
+        \ '%E%f: line %l\, col %c\, Error - %m,' .
+        \ '%W%f: line %l\, col %c\, Warning - %m'
+
+    let loclist = SyntasticMake({
+        \ 'makeprg': makeprg,
+        \ 'errorformat': errorformat,
+        \ 'defaults': {'type': 'W'},
+        \ 'returns': [0, 1] })
+
+    return loclist
+endfunction
+
+call g:SyntasticRegistry.CreateAndRegisterChecker({
+    \ 'filetype': 'javascript',
+    \ 'name': 'xo'})
+
+let &cpo = s:save_cpo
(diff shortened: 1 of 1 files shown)


lcd047:
There seems to be a recent trend of writing various wrappers and frontends to `eslint`, so I tried to make the `eslint` checker accommodate them easier.  Please upgrade syntastic to commit c34ea89 (or later), and set:
```vim
let g:syntastic_javascript_eslint_generic = 1
let g:syntastic_javascript_eslint_exec = 'xo'
let g:syntastic_javascript_eslint_args = '--compact'
let g:syntastic_javascript_checkers = ['eslint']
```

arthurvr:
Cool, thanks!

sindresorhus:
> so I tried to make the eslint checker accommodate them easier

I don't really see the point of this. You're now putting the burden on the user instead of having a tiny bit of duplicated code in the repo. I would rather have made the ESLint plugin subclass-able or something, so it could be reused in the XO syntax checker. The whole point of XO over ESLint is to remove all config. Now I will have to document how the user has to manually configure the ESLint plugin, instead of just choosing the XO plugin, and be done with it...

lcd047:
> You're now putting the burden on the user

Yes, whenever reasonably possible.  Syntastic has currently 187 checkers for 94 different filetypes, spanning almost as many different languages.  I'm currently the only active maintainer left, and I know little more than the name about most of these languages.  I'm also doing all this for free (actually, also for fun, but having fun doesn't increase my amount of available resources), in my spare time.

> instead of having a tiny bit of duplicated code in the repo.

And pretty soon end up with ~10 different copies of the same checker, subtly different to one another, and going out of synch with each other.  Thanks, but no thanks.

> I would rather have made the ESLint plugin subclass-able or something, so it could be reused in the XO syntax checker.

That's exactly what the generic thing does, except it's the other way around: the XO checker is a subclass of the ESLint one.

> The whole point of XO over ESLint is to remove all config.

This is your goal, not syntastic's (or mine, for what that's worth).

> Now I will have to document how the user has to manually configure the ESLint plugin, instead of just choosing the XO plugin, and be done with it...

Sometimes you can change the world according to your views.  Most of the time it's the other way around.

lcd047:
@sindresorhus On a side note: you might want to take a look at #1553 for a different approach to the same problem.

sindresorhus:
> Syntastic has currently 187 checkers for 94 different filetypes, spanning almost as many different languages. I'm currently the only active maintainer left, and I know little more than the name about most of these languages. I'm also doing all this for free (actually, also for fun, but having fun doesn't increase my amount of available resources), in my spare time. The result is what it is.

Totally understandable. That's a crazy amount of checkers. Seems like a good oppertunity to support linter plugins. In Atom/Sublime, every linter is just another package and the only thing the linter core maintainers have to care about is the linter core itself, not every imaginable plugin. Plugins are maintained by community member with knowledge of the different ecosystems. I would be happy to maintain such plugin for XO.

> That's exactly what the generic thing does, except it's the other way around: the XO checker is a subclass of the ESLint one.

That's exactly what I asked for. I might not have been clear enough. But, I would just want

```
let g:syntastic_javascript_eslint_generic = 1
let g:syntastic_javascript_eslint_exec = 'xo'
let g:syntastic_javascript_eslint_args = '--compact'
let g:syntastic_javascript_checkers = ['eslint']
```

to be *the* XO checker in this repo. So when the user choses the `xo` checker in Syntastic, it would just use the ESLint checker with the following lines. I understand you don't want to maintain almost identical full checkers, but a shortcut like this would make it a lot easier for users. I would be happy to maintain those 4 lines in this repo (they're probably never going to change).

> This is your goal, not syntastic's (or mine, for what that's worth).

I know. Just there to help you understand why I'm figting for this.

lcd047:
> Seems like a good oppertunity to support linter plugins.

It's [already done](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).  Except Vim makes it more of a pain in the rear than Python. _shrug_

> I would be happy to maintain such plugin for XO.

All you have to do is put the plugin in the PR above a file `syntax_checkers/javascript/xo.vim`.  You'll probably also want to clean it up a little (f.i. the `IsAvailable()` function can be omitted).

sindresorhus:
Just for posterity, I ended up making an external checker: https://github.com/sindresorhus/vim-xo

PR #1567, opened 2015-10-02
API Blueprint should use drafter

kylef:
This is currently using the `snowcrash` command line tool which has been [non-existent since March](https://github.com/apiaryio/snowcrash/releases/tag/v1.0.0). and replaced in favour of [Drafter](https://github.com/apiaryio/drafter).

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -8,7 +8,7 @@ let g:loaded_syntastic_registry = 1
 let s:_DEFAULT_CHECKERS = {
         \ 'actionscript':  ['mxmlc'],
         \ 'ada':           ['gcc'],
-        \ 'apiblueprint':  ['snowcrash'],
+        \ 'apiblueprint':  ['drafter'],
         \ 'applescript':   ['osacompile'],
         \ 'asciidoc':      ['asciidoc'],
         \ 'asm':           ['gcc'],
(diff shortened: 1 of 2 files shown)


kylef:
Screenshot of Syntastic + drafter with the [API Blueprint](https://github.com/kylef/apiblueprint.vim) plugin:

lcd047:
Does `drafter` have the exact same command line arguments as the old `snowcrash`?  Does it produce error messages in the the exact same format?

kylef:
@lcd047 Yes, it takes both the `-u` and `-l` arguments that this script is using and produces the same output as snowcrash did.

```
$ drafter --help
usage: drafter [options] ... <input file>

API Blueprint Parser
If called without <input file>, 'drafter' will listen on stdin.

options:
  -o, --output          save output AST into file (string [=])
  -f, --format          output format of the AST (yaml|json) (string [=yaml])
  -t, --type            type of the AST (refract|ast) (string [=refract])
  -s, --sourcemap       export sourcemap AST into file (string [=])
  -h, --help            display this help message
  -v, --version         print Drafter version
  -l, --validate        validate input only, do not print AST
  -u, --use-line-num    use line and row number instead of character index when printing annotation
```

kylef:
Thanks @lcd047.

lcd047:
Normally I would handle this by renaming the `exec` in the old checker, for backward compatibility.  However, since `snowcrash` doesn't care about backward compatibility, there doesn't seem to be much point in doing it in syntastic.  Merged, thank you.

PR #1573, opened 2015-10-14
Update exit codes for drafter

kylef:
`drafter` may return underlying errors from `snowcrash` it's internal library for parsing for various reasons.

I've updated the list in syntastic to include any blueprint error.

List of exit codes can be found at https://github.com/apiaryio/snowcrash/blob/master/src/SourceAnnotation.h#L100-L106. I've intentionally left out ApplicationError (1) as this would be an application error and not a blueprint error.

Thanks :octocat:.

### DIFF
--- a/syntax_checkers/apiblueprint/drafter.vim
+++ b/syntax_checkers/apiblueprint/drafter.vim
@@ -34,7 +34,7 @@ function! SyntaxCheckers_apiblueprint_drafter_GetLocList() dict
         \ 'makeprg': makeprg,
         \ 'errorformat': errorformat,
         \ 'defaults': {'bufnr': bufnr('')},
-        \ 'returns': [0, 2] })
+        \ 'returns': [0, 2, 3, 4] })
 
     for e in loclist
         let matches = matchlist(e['text'], '\v^(.+); line (\d+), column (\d+) - line (\d+), column (\d+)$')


lcd047:
Leaving out anything but 0 and 2 is intentional.  Syntastic doesn't produce any useful output for `ModelError` and `MSONError`, and you want to know there's a problem.

kylef:
I don't think that's correct @lcd047.

The syntastic checker already prints all warnings and errors, that would include model errors and mson errors.

For example, see my attachment:

screen shot 2015-10-13 at 22 24 57

With these changes, syntastic is capable of showing these errors:

screen shot 2015-10-13 at 22 26 08

Without them it will error incorrectly:

screen shot 2015-10-13 at 22 27 38

lcd047:
Then let's put it this way: at the time I looked at the sources to make sure the `errorformat` works, I only looked at return code 2.  IIRC return codes 3 and 4 can produce errors in other formats, and they would be silently discarded if syntastic accepted 3 and 4 as valid.  Back then, the checker wasn't even named `drafter`.

Now, if you look at the sources and guarantee `drafter` returns errors in the same format for exit codes 3 and 4 as it does for 2, I'll take your word for it.  But somebody has to check that, and I'm afraid I don't have time to do it myself right now. _shrug_

pksunkara:
> Now, if you look at the sources and guarantee drafter returns errors in the same format for exit codes 3 and 4 as it does for 2, I'll take your word for it.

I am Pavan. I work at Apiary which involves developing the API Blueprint parser. I am the guy who added those error codes. Yes, they are the same format as the error code 2. @kylef is my colleague.

lcd047:
Ok then, merged, thank you.

PR #1579, opened 2015-10-24
Add support for Icarus Verilog compiler

Psidium:
For my CS class I needed to use this compiler for Verilog and my favourite editor/syntax checker didn't have support for it, so, instead of doing my assignment for my class, I've added support for it :smile:, now I'm uploading it back to the community. 
This new syntax checker is based on `verilator.vim`.
I've read `contributing.md` and I don't see any errors in my patch.

Also, I'd like to note that this task got SO easy once I discovered `let g:syntastic_debug = 3`, congratulation on such an epic project like this :heart_eyes:

### DIFF
--- a/syntax_checkers/verilog/iverilog.vim
+++ b/syntax_checkers/verilog/iverilog.vim
@@ -9,22 +9,22 @@ if exists('g:loaded_syntastic_verilog_iverilog_checker')
 endif
 let g:loaded_syntastic_verilog_iverilog_checker = 1
 
-if !exists('g:syntastic_verilog_compiler_options')
-    let g:syntastic_verilog_compiler_options = '-Wall'
-endif
-
 let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_verilog_iverilog_GetLocList() dict
-    if !exists('g:syntastic_verilog_compiler')
-        let g:syntastic_verilog_compiler = self.getExec()
-    endif
-    return syntastic#c#GetLocList('verilog', 'iverilog', {
-        \ 'errorformat':
+    let makeprg = self.makeprgBuild({
+                \ 'args_before': '-t null',
+                \ 'args': '-Wall' })
+
+    let errorformat =
         \     '%f:%l: %trror: %m,' .
-        \     '%f:%l: %tarning: %m',
-        \ 'main_flags': '-t null' })
+        \     '%f:%l: %tarning: %m,' .
+        \     '%E%f:%l:      : %m,' .
+        \     '%W%f:%l:        : %m,' .
+        \     '%f:%l: %m'
+
+    return SyntasticMake({'makeprg': makeprg, 'errorformat': errorformat})
 endfunction
 
 call g:SyntasticRegistry.CreateAndRegisterChecker({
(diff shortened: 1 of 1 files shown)


Psidium:
Noted just now, do I need to edit the wiki as well?

lcd047:
Thank you for your patch.

Is there any particular reason why you used `syntastic#c#GetLocList()` instead of the more common (and [documented](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide)) `makeprgBuild()`?  Do you understand the difference?  Are you aware of the limitations of `syntastic#c#GetLocList()`?

Psidium:
Hello!
This function was used because `verilator.vim` used it, and since it worked for me, I didn't see a point in changing to `makeprgBuild()`. Would you like me to change it?

lcd047:
> This function was used because `verilator.vim` used it, and since it worked for me, I didn't see a point in changing to `makeprgBuild()`. 

Perhaps follow the docs next time, or at least ask what's going on?

> Would you like me to change it?

With it, your checker conflicts with `verilator`.  Among other things, both your checker and `verilator` define `g:syntastic_verilog_compiler`, with conflicting values.

For historic reasons, at most one checker per filetype can use `syntastic#c#GetLocList()`, and for `verilog` that position is already taken by `verilator`.  If you must have two checkers using `syntastic#c#GetLocList()` for the same filetype, you can look at how `fortran/gfortran` does it: by having both checkers share the same syntastic code.  This, of course, should be avoided whenever possible.

Psidium:
Hi, 
I've updated my patch to use `makeprgBuild()` and this change has make the code so much cleaner and concise, sorry for not having used it at first.
Can you please see if it's ok now?
Thanks

lcd047:
Looking at the [sources][src], I suggest a slightly different `errorformat`:
```
let &errorformat =
    \ '%W%f:%l: %tarning: %m,' .
    \ '%C%f:%l:        : %m,' .
    \ '%E%f:%l: %trror: %m,' .                                                                                                  
    \ '%C%f:%l:      : %m,' .
    \ '%f:%l: %m'
```
If you do that you'll also need `compressWhitespace` postprocessing.

Also, please add a licence to your checker.

[src]: https://github.com/steveicarus/iverilog

Psidium:
Hi, thank you for your suggestion.
I've added MIT to the license, it's a license I'm comfort with, but if it's needed, I can change it to another one without a problem.

If I change `errorformat` to your proposed change, I get unwanted behaviour.
Sometimes the line (`%l`) can differ even inside the same error. for example: 
```text
iverilog -o main main.v
main.v:42: warning: Port 9 (ab_result) of fsm expects 8 bits, got 5.
main.v:42:        : Padding (signed) 3 high bits of the port.
main.v:29: error: initCount is not a valid l-value in main.
main.v:8:      : initCount is declared here as wire.
1 error(s) during elaboration.
```
The line starting with `main.v:8` is indeed the same error as the line starting with `main.v:29: error` but it point somewhere else. If I use your suggestion of `errorformat`, the line `main.v:8:` is ignored (although the message that was supposed to appear at line `8` appears at line `29` appended with the original message).

It is needed that syntastic points to the both lines, and the only solution I found was to treat as different errors.

If you have another suggestion about how to preserve both lines while pointing to the same error, I'm glad to implement it.

Thanks

lcd047:
Ok, merged, thank you.

PR #1586, opened 2015-10-29
Add CSS Stylelint checker

pixelastic:
Hello,

I've added a syntax checker for [stylelint](http://stylelint.io/), a CSS linter.

As far as I can tell, it produces two different outputs, based on the severity of the issue. Basically it "crashes" if the CSS cannot be parsed, and display linting results otherwise. Both are handled by this PR.

As usual, `errorformat` still looks a bit like voodoo to me so maybe things can be done a better way.

### DIFF
--- a/syntax_checkers/css/stylelint.vim
+++ b/syntax_checkers/css/stylelint.vim
@@ -1,0 +1,46 @@
+"============================================================================
+"File:        stylelint.vim
+"Description: Syntax checking plugin for syntastic.vim using `stylelint`
+"             (https://github.com/stylelint/stylelint).
+"Maintainer:  Tim Carry <tim at pixelastic dot com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"
+"============================================================================
+
+if exists('g:loaded_syntastic_css_stylelint_checker')
+    finish
+endif
+let g:loaded_syntastic_css_stylelint_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_css_stylelint_GetLocList() dict
+    let makeprg = self.makeprgBuild({ })
+    " Example output:
+    " CssSyntaxError: /path/to/file.css:2:11: Missed semicolon
+    " or
+    " 2:11    Expected single space after ":" (declaration-colon-space-after)
+    let errorformat = 
+          \ 'CssSyntaxError: %f:%l:%c: %m,' .
+          \ '%\m%l\:%c%\s%#%m,'
+
+    return SyntasticMake({
+        \ 'makeprg': makeprg,
+        \ 'errorformat': errorformat,
+        \ 'defaults': {'bufnr': bufnr('')} })
+endfunction
+
+call g:SyntasticRegistry.CreateAndRegisterChecker({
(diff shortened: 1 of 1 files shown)


lcd047:
Thank you.  Merged, with this patch on top: c596b64.

pixelastic:
Thanks.

I'm not sure to understand what your patch is doing. Could you elaborate (so I can submit better PRs next time)?

What I understand is that you request an output as JSON (I did not even known that option existed...) then check in a preprocess if it "fails" or not, then you return the error in an easy, `:` delimited format so the `errorformat` is easier to handle. Am I correct?

What does the `'returns': [0, 1, 2]` means? And I think the preprocess should warn about `css/stylelint` and not `javascript/flow` :)

lcd047:
> I'm not sure to understand what your patch is doing.

It mainly replace parsing text output with parsing JSON.  The rest are just cosmetic details.

> Could you elaborate (so I can submit better PRs next time)?

Oh, you didn't do anything "wrong", it's just that the JSON output is more precise than the text one: it allows distinguishing between errors (messages with severity 2) and warnings (severity 1).

> What I understand is that you request an output as JSON (I did not even known that option existed...)

Yeah, I found out about it from looking at the sources.  `stylelint --help` produces a short help, but `stylelint -h` and `stylelint -?`, creatively, don't.

> then check in a preprocess if it "fails" or not, then you return the error in an easy,  `:` delimited format so the `errorformat` is easier to handle. Am I correct?

Not really.  Syntastic has two ways of parsing output from checkers: using `errorformat`, and manual parsing.  In principle the `errorformat` way is easier, but `errorformat` has various limitations and [misfeatures](https://groups.google.com/forum/#!topic/vim_dev/pTKmZmouhio).  Syntastic shouldn't have used it to begin with, but that's way too late to change now.  I suppose choosing it made sense at the beginning of the project, since `errorformat` is what Vim offers specifically for parsing errors, and the "obvious" alternative, using regexps, is more complicated than it should be because Vim regexps don't allow named subexpressions.

To overcome limitations in `errorformat`, syntastic has added [`preprocess` and `postprocess` functions](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#syntasticmakeoptions).  Preprocess functions take the raw output of the checker on input, and munge it before feeding it back to `errorformat`.  The initial idea was that the raw output could be a pain in the rear to parse with `errorformat` alone, but sometimes minor adjustments could bring it to sanity.  Then it turned out that preprocess functions were perfect for doing things like parsing JSON or XML, which couldn't be done with `errorformat`.  So they became the "go to" solution for general manual parsing.  Most of these functions could easily return `loclist`s instead of something that can be trivially parsed with `errorformat`.  But that would require a redesign of major parts of syntastic, and it would break compatibility with third-party checkers.  Plus, the existing modularity has its advantages: a few of these functions can (and are) used by multiple checkers, and the structure of the data passed around is mostly trivial.  And the inefficiency of feeding the result of manual parsing to `errorformat` is negligible compared to the inefficiency of running external programs.

Postprocess functions are less interesting: they take error messages produced by `errorformat` and munge them.  This is useful for things like changing character encoding, and decoding XML entities.  They too can be (and are) re-used by several checkers.

> What does the `'returns': [0, 1, 2]` means?

It's explained in the [guide](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#syntasticmakeoptions): it checks the exit code of the checker.  Syntastic will bail out if the checker returns anything else than the list of values pointed to by `returns`.  The point is to let the user know something went wrong.  The result is not always great, but it's more than nothing.

> And I think the preprocess should warn about  css/stylelint  and not  javascript/flow  :)

Right, cut & waste error. :) Should be fixed in 630169b.

PR #1598, opened 2015-11-13
Improve clang diagnostics, support compilation database

Lekensteyn:
Two fixes (in one patch):
___
Propagate informational messages from clang, this can be useful for
syntax errors:

    packet-ssl-utils.c:6799:4: error: expected '}'
    packet-ssl-utils.c:6778:1: note: to match this '{'
___
Clearing g:syntastic_c_clang_check_post_args as documented on the wiki
does not work when colorful diagnostics are enabled (the output is not
recognized). Fix this by adding a new option for the compilation
database path that preserves the formatting options.
___
If this patch is approved, the wiki must also be updated:
https://github.com/scrooloose/syntastic/wiki/C%3A---clang_check

### DIFF
--- a/syntax_checkers/c/clang_check.vim
+++ b/syntax_checkers/c/clang_check.vim
@@ -22,24 +22,34 @@ if !exists('g:syntastic_c_clang_check_sort')
     let g:syntastic_c_clang_check_sort = 1
 endif
 
+if !exists('g:syntastic_c_clang_check_database')
+    let g:syntastic_c_clang_check_database = ''
+endif
+
 let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_c_clang_check_GetLocList() dict
+    if g:syntastic_c_clang_check_database == ''
+        let l:extra_args = '-- ' .
+        \   syntastic#c#ReadConfig(g:syntastic_clang_check_config_file)
+    else
+        let l:extra_args = '-p=' . g:syntastic_c_clang_check_database
+    endif
     let makeprg = self.makeprgBuild({
         \ 'post_args':
-        \   '-- ' .
-        \   syntastic#c#ReadConfig(g:syntastic_clang_check_config_file) . ' ' .
-        \   '-fshow-column ' .
-        \   '-fshow-source-location ' .
-        \   '-fno-caret-diagnostics ' .
-        \   '-fno-color-diagnostics ' .
-        \   '-fdiagnostics-format=clang' })
+        \   '-extra-arg=-fshow-column ' .
+        \   '-extra-arg=-fshow-source-location ' .
+        \   '-extra-arg=-fno-caret-diagnostics ' .
+        \   '-extra-arg=-fno-color-diagnostics ' .
+        \   '-extra-arg=-fdiagnostics-format=clang ' .
+        \   l:extra_args })
 
     let errorformat =
         \ '%E%f:%l:%c: fatal error: %m,' .
         \ '%E%f:%l:%c: error: %m,' .
         \ '%W%f:%l:%c: warning: %m,' .
+        \ '%W%f:%l:%c: note: %m,' .
         \ '%-G%\m%\%%(LLVM ERROR:%\|No compilation database found%\)%\@!%.%#,' .
         \ '%E%m'
 
--- a/syntax_checkers/c/clang_tidy.vim
+++ b/syntax_checkers/c/clang_tidy.vim
@@ -22,24 +22,34 @@ if !exists('g:syntastic_c_clang_tidy_sort')
     let g:syntastic_c_clang_tidy_sort = 1
 endif
 
+if !exists('g:syntastic_c_clang_tidy_database')
+    let g:syntastic_c_clang_tidy_database = ''
+endif
+
 let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_c_clang_tidy_GetLocList() dict
+    if g:syntastic_c_clang_tidy_database == ''
+        let l:extra_args = '-- ' .
+        \   syntastic#c#ReadConfig(g:syntastic_clang_tidy_config_file)
+    else
+        let l:extra_args = '-p=' . g:syntastic_c_clang_tidy_database
+    endif
     let makeprg = self.makeprgBuild({
         \ 'post_args':
-        \   '-- ' .
-        \   syntastic#c#ReadConfig(g:syntastic_clang_tidy_config_file) . ' ' .
-        \   '-fshow-column ' .
-        \   '-fshow-source-location ' .
-        \   '-fno-caret-diagnostics ' .
-        \   '-fno-color-diagnostics ' .
-        \   '-fdiagnostics-format=clang' })
+        \   '-extra-arg=-fshow-column ' .
+        \   '-extra-arg=-fshow-source-location ' .
+        \   '-extra-arg=-fno-caret-diagnostics ' .
+        \   '-extra-arg=-fno-color-diagnostics ' .
+        \   '-extra-arg=-fdiagnostics-format=clang ' .
+        \   l:extra_args })
 
     let errorformat =
         \ '%E%f:%l:%c: fatal error: %m,' .
         \ '%E%f:%l:%c: error: %m,' .
         \ '%W%f:%l:%c: warning: %m,' .
+        \ '%W%f:%l:%c: note: %m,' .
         \ '%-G%\m%\%%(LLVM ERROR:%\|No compilation database found%\)%\@!%.%#,' .
         \ '%E%m'
 


Lekensteyn:
Thinking about it, it is possible to support the extra config file when using compilation databases, every line must then be prefixed with `-extra-arg=`. My vim script-fu is not that good, so if that behavior is desired, I would need some guidance :)

lcd047:
> Propagate informational messages from clang, this can be useful for
 syntax errors:

This is wrong (or at least your attempted implementation of it is).  Syntastic only supports error levels 
`error` and `warning`.

> Clearing g:syntastic_c_clang_check_post_args as documented on the wiki
 does not work when colorful diagnostics are enabled (the output is not
 recognized).

Please provide evidence of your claim.  Open your test file, set `g:syntastic_debug` to 3, run the checker, run `:mes`, and post the output.

Lekensteyn:
>> Propagate informational messages from clang, this can be useful for syntax errors:
>
> This is wrong (or at least your attempted implementation of it is).  Syntastic only supports error levels `error` and `warning`.

I saw other checkers using the `%I` type (bison). What I observe with this change is a blue marker (but no message unfortunately). I am OK with adding `%W` instead, but maybe that could be a bit more noisy.

>> Clearing g:syntastic_c_clang_check_post_args as documented on the wiki
>> does not work when colorful diagnostics are enabled (the output is not
>> recognized).
>
> Please provide evidence of your claim.  Open your test file, set `g:syntastic_debug` to 3, run the checker, run `:mes`, and post the output.

I set it to `-p=/path/to/directory/with/compile_commands.json`, but the output without that option would be the same. Here is the relevant output showing ANSI color escapes:
```
syntastic: 35.580465: CacheErrors: Invoking checker: c/clang_check
syntastic: 35.581132: SyntasticMake: called with options: {'errorformat': '%E%f:%l:%c: fatal error: %m,%E%f:%l:%c: error: %m,%W%f:%l:%c: warning: %m,%-G%\m%\%%(LLVM ERROR:%\|No co
mpilation database found%\)%\@!%.%#,%E%m', 'defaults': {'bufnr': 1}, 'makeprg': 'clang-check epan/dissectors/packet-imap.c -p=/tmp/wsbuild', 'returns': [0, 1]}
syntastic: 35.764641: checker output: ['^[[1m/tmp/wireshark/epan/dissectors/packet-imap.c:378:4: ^[[0m^[[0;1;31merror: ^[[0m^[[1mexpected ''}''^[[0m', ' */', '^[[0;1;32m   ^', '^[
[0m^[[1m/tmp/wireshark/epan/dissectors/packet-imap.c:362:1: ^[[0m^[[0;1;30mnote: ^[[0mto match this ''{''^[[0m', '{', '^[[0;1;32m^', '^[[0m1 error generated.', 'Error while proces
sing /home/peter/projects/wireshark/epan/dissectors/packet-imap.c.', '']
syntastic: 35.765980: raw loclist: []
syntastic: 35.766370: getLocList: checker c/clang_check returned 1
```

To verify that these options are really needed, I put the options (that would be disabled) in the post_args variable of my vimrc and then things started to work again. But that sounds like a bug in the plugin which is why this patch is proposed.

lcd047:
> I saw other checkers using the `%I` type (bison).

Once you understand the code in the `bison` checker, you'll understand that type `I` is not actually used.

> the output without that option would be the same.

Well, then you might want to report this to the Clang project.  `clang-check` should never output color codes when the output is not a terminal -- and that's the case when `clang-check` is run from Vim.

Lekensteyn:
Updated the patch to use `%W` instead of `%I` and to set `TERM=dumb` to use no color by default. The `-post-arg=` changes are still needed when the compile flags contain `-fcolor-diagnostics`. See the updated commit message for more details.

lcd047:
Actually, the `TERM` part sounds useful: commit 7d62700.

Lekensteyn:
What about the other changes to take the "note:" lines from the clang output and make the output more useful? What is the primary concern you have with the patch/idea (if any)?

lcd047:
> What is the primary concern you have with the patch/idea (if any)?

Oh, I have a few.  Adding workarounds to syntastic for flaws in external programs is bad.  Incompatible changes are bad.  Features that can't be turned off are bugs.  Adding new options to checkers is a pain in the rear in the long run.  I could go on, but I doubt you care to know more. :)

I'm not really convinced informational messages would be anything but annoying to most people.  I know they would be to me.  And nobody has asked for them so far. _shrug_

Lekensteyn:
Alright, the `-fcolor-diagnostics` option is the only option that is overridden, so a workaround in vimrc should be a good compromise. I'll leave the informational messages for now. The blue marker can be helpful, in case I really miss it I will re-open this issue :)

Here is a diff for `git@github.com:scrooloose/syntastic.wiki.git`, I cannot push to the wiki and editing all pages manually does not sound nice (I have already modified the clang-check page). Hopefully it is written in a good shape:

    diff --git a/C++:---clang_check.md b/C++:---clang_check.md
    index 757c31d..c71f023 100644
    --- a/C++:---clang_check.md
    +++ b/C++:---clang_check.md
    @@ -23,6 +23,15 @@ let g:syntastic_cpp_clang_check_post_args = ""
     ```
     (configuration files pointed to by `g:syntastic_clang_check_config_file` are then ignored).
     
    +Should the compilation database be located in a directory separate from
    +the source tree, then you can pass the `-p=PATH` option. If the compile
    +commands contain `-fcolor-diagnostics`, then these must be turned off
    +or you will get no diagnostics:
    +```vim
    +let g:syntastic_cpp_clang_check_post_args =
    +\ "-p=/tmp/builddir -extra-arg=-fno-color-diagnostics"
    +```
    +
     [0]: http://clang.llvm.org/docs/ClangCheck.html
     [1]: http://clang.llvm.org/docs/JSONCompilationDatabase.html
     [2]: http://cmake.org/
    diff --git a/C++:---clang_tidy.md b/C++:---clang_tidy.md
    index c3729f4..c66ffbb 100644
    --- a/C++:---clang_tidy.md
    +++ b/C++:---clang_tidy.md
    @@ -29,6 +29,15 @@ let g:syntastic_cpp_clang_tidy_post_args = ""
     ```
     (configuration files pointed to by `g:syntastic_clang_tidy_config_file` are then ignored).
     
    +Should the compilation database be located in a directory separate from
    +the source tree, then you can pass the `-p=PATH` option. If the compile
    +commands contain `-fcolor-diagnostics`, then these must be turned off
    +or you will get no diagnostics:
    +```vim
    +let g:syntastic_cpp_clang_tidy_post_args =
    +\ "-p=/tmp/builddir -extra-arg=-fno-color-diagnostics"
    +```
    +
     [0]: http://clang.llvm.org/extra/clang-tidy.html
     [1]: http://clang.llvm.org/docs/JSONCompilationDatabase.html
     [2]: http://cmake.org/
    diff --git a/C:---clang_check.md b/C:---clang_check.md
    index 45f2b66..fa3fe34 100644
    --- a/C:---clang_check.md
    +++ b/C:---clang_check.md
    @@ -28,7 +28,8 @@ the source tree, then you can pass the `-p=PATH` option. If the compile
     commands contain `-fcolor-diagnostics`, then these must be turned off
     or you will get no diagnostics:
     ```vim
    -let g:syntastic_c_clang_check_post_args = "-p=/tmp/builddir -extra-arg=-fno-color-diagnostics"
    +let g:syntastic_c_clang_check_post_args =
    +\ "-p=/tmp/builddir -extra-arg=-fno-color-diagnostics"
     ```
     
     [0]: http://clang.llvm.org/docs/ClangCheck.html
    diff --git a/C:---clang_tidy.md b/C:---clang_tidy.md
    index 650cd86..37cfc45 100644
    --- a/C:---clang_tidy.md
    +++ b/C:---clang_tidy.md
    @@ -29,6 +29,15 @@ let g:syntastic_c_clang_tidy_post_args = ""
     ```
     (configuration files pointed to by `g:syntastic_clang_tidy_config_file` are then ignored).
     
    +Should the compilation database be located in a directory separate from
    +the source tree, then you can pass the `-p=PATH` option. If the compile
    +commands contain `-fcolor-diagnostics`, then these must be turned off
    +or you will get no diagnostics:
    +```vim
    +let g:syntastic_c_clang_tidy_post_args =
    +\ "-p=/tmp/builddir -extra-arg=-fno-color-diagnostics"
    +```
    +
     [0]: http://clang.llvm.org/extra/clang-tidy.html
     [1]: http://clang.llvm.org/docs/JSONCompilationDatabase.html
     [2]: http://cmake.org/

lcd047:
> If the compile
> +commands contain `-fcolor-diagnostics`, then these must be turned off
> +or you will get no diagnostics:

So, your problem is actually that if you turn on `-fcolor-diagnostics` then you also need to turn it off.  How about not putting it there to begin with.  Not _all_ possible flag combinations are supported, and syntastic explicitly doesn't support user changes that affect output format.  I'm closing this, sorry.

Lekensteyn:
I expect syntastic "just to work" given common flags, apparently that is too much asked.

lcd047:
Syntastic parses the output of third-party programs.  Syntastic has to make assumptions about them.  One of the fundamental assumptions is that the user doesn't mess with options that affect said programs output.  Also, these are moving targets, "just work" is not a goal for syntastic.  "Hopefully works, given reasonable conditions and user's cooperation" is a more realistic goal.  Sorry about that.

PR #1599, opened 2015-11-15
Add ansible syntax checker

erikzaadi:
Luckily, `ansible-lint` has a parameter to output `pep8` compatible output, so I was able to base most of this plugin on the `pep8` syntax checker.

### DIFF
--- a/syntax_checkers/ansible/ansible-lint.vim
+++ b/syntax_checkers/ansible/ansible-lint.vim
@@ -1,0 +1,48 @@
+"============================================================================
+"File:        ansible-lint.vim
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Erik Zaadi <erik.zaadi at gmail dot com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"
+"============================================================================
+
+if exists('g:loaded_syntastic_ansible_lint_checker')
+    finish
+endif
+let g:loaded_syntastic_ansible_lint_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_ansible_lint_GetLocList() dict
+    let makeprg = self.makeprgBuild({'args': '-p'})
+
+    let errorformat = '%f:%l: %m'
+
+    let env = syntastic#util#isRunningWindows() ? {} : { 'TERM': 'dumb' }
+
+    let loclist = SyntasticMake({
+        \ 'makeprg': makeprg,
+        \ 'errorformat': errorformat,
+        \ 'env': env})
+
+    for e in loclist
+        let e['type'] = e['text'] =~? '^W' ? 'W' : 'E'
+    endfor
+
+    return loclist
+endfunction
+
+call g:SyntasticRegistry.CreateAndRegisterChecker({
+    \ 'filetype': 'ansible',
+    \ 'name': 'lint', 
+    \ 'exec': 'ansible-lint'})
+
+let &cpo = s:save_cpo
+unlet s:save_cpo
+
+" vim: set sw=4 sts=4 et fdm=marker:


lcd047:
Thank you.  Merged, with this patch on top: 721a42d.

erikzaadi:
Thanks @lcd047 !

badnetmask:
Not sure if I did this wrong, but as far as I understand I need to set "# vim:ft=ansible:" at the end of the file to make the plugin work, however when I do:
a) I loose colored visualization
b) Syntax check itself doesn't work (running ansible-lint manually on a file gives me an error, but saving it from vim doesn't do anything)
I am sure syntastic works fine because I use it daily for puppet linting.
Any thoughts?

erikzaadi:
@badnetmask I'm testing now with the patched changes, will update here.

Does that occur when you do `set filetype=ansible` instead?
I use it with https://github.com/chase/vim-ansible-yaml , so no need for the filetype definition per file.

erikzaadi:
@badnetmask Worked on my box.
Could you try `let g:syntastic_debug = 1`
Then open an ansible-yaml file, `:SyntasticCheck`
Then `:mes` to see the error message

lcd047:
@badnetmask Yes, you need to set the `filetype` of the files you want to check to `ansible`.  This is not a builtin filetype for Vim, so yes, you're likely to lose syntax highlighting, indent settings, and a few other things when you do that.  Normally there should be some kind of plugin to define a new `ansible` filetype and take care of all details, but such a plugin would have nothing to do with syntastic.  The syntastic checker should work if you set the filetype to `ansible`, it doesn't really care about anything else.

badnetmask:
@erikzaadi I don't use vim-ansible-yaml, so I rely on vim to figure out syntax highlight. I am guessing it thinks it's a yaml file, as I use yml extension on my playbooks.

This is what I see on the debug output when I run the check without "ft=ansible":
syntastic: 11.027092: CacheErrors: Checker yaml/jsyaml is not available
syntastic: 11.027225: CacheErrors: no checkers available for yaml

And this when I set "ft=ansible":
syntastic: 9.481946: CacheErrors: Invoking checker: ansible/ansible_lint
syntastic: 9.483271: SyntasticMake: called with options: {'subtype': 'Style', 'errorformat': '%f:%l: [ANSIBLE%n] %m', 'defaults': {'type': 'E'}, 'makeprg': 'an
sible-lint -p run.yaml', 'env': {'TERM': 'dumb'}, 'returns': [0, 2]}
syntastic: 9.696455: getLocList: checker ansible/ansible_lint returned 2

Apparently the error is the '-p'.
$ ansible-lint -p run-puppet.yaml 
Usage: ansible-lint playbook.yml

ansible-lint: error: no such option: -p

$ ansible-lint --version
ansible-lint 2.0.3

By the way this is the lint output when I run without -p:
ansible.errors.AnsibleYAMLValidationFailed: Syntax Error while loading YAML script, run-puppet.yaml
Note: The error may actually appear before this position: line 4, column 3

  user root
  gather_facts: no
  ^

By the way, the only positive thing of installing vim-ansible-yaml is that now I get syntax highlighting when I set "ft=ansible", however the file still gets detected as yaml when the tag is not set.

erikzaadi:
@badnetmask 
Bah, `ansible-lint` added that in a later version, I have it in 2.1.3

https://github.com/willthames/ansible-lint/commit/1ceaae1eddb32d0e4068d9d1787e10f692fdc14a

Seems like it was added in 2.0.4

lcd047:
Great, `-p` was added in 2.0.4.  Ok, commit 13c8e75 checks for the right `ansible-lint` version.

badnetmask:
So I'm guessing I've been cut off. :)

lcd047:
@badnetmask You can always upgrade to a more recent `ansible-lint`.

badnetmask:
@lcd047 I'm using Fedora 23, which still has old package. I've opened a bug requesting an update. Thanks.

badnetmask:
OK, I am now using ansible-lint from github, and syntastic is not working as I would expect. I use it daily to validate Puppet code, and when I save something bad I get a new vim window with the errors, while with an ansible file all I see is a red bad that appears quickly on the screen, then disappears, and it gives me back the bash prompt.

Running on debug mode, manually, this is what I see:
syntastic: 5.288321: CacheErrors: Invoking checker: ansible/ansible_lint
syntastic: 5.288885: SyntasticMake: called with options: {'subtype': 'Style', 'errorformat': '%f:%l: [ANSIBLE%n] %m', 'defaults': {'type': 'E'}, 'makeprg': 'an
sible-lint -p run-puppet.yml', 'env': {'TERM': 'dumb'}, 'returns': [0, 2]}
syntastic: error: checker output:
<big traceback>
ansible.errors.AnsibleYAMLValidationFailed: Syntax Error while loading YAML script, run-puppet.yml
Note: The error may actually appear before this position: line 4, column 3
  user root
  gather_facts: no
  ^
syntastic: error: checker ansible/ansible_lint returned abnormal status 1

lcd047:
@badnetmask You removed the `-p`.

lcd047:
Congratulations: you found a bug in `ansible-lint`.  Syntax errors are not reported in the parseable format, regardless of the `-p` option.  You know how to handle this, I presume. :)

badnetmask:
I don't. I thought @erikzaadi would. :)

lcd047:
@badnetmask Post a detailed bug report to the `ansible-lint` issue tracker?

erikzaadi:
@badnetmask I gotta admit, I have ansible files set to `filetype=yaml.ansible` so the yaml parser validates it before `ansible-lint`.

But I'll do another topic branch that addresses this with only using `ansible-lint`

badnetmask:
@erikzaadi even setting ft=yaml.ansible it still ignore syntax errors, so sounds more like what @lcd047 said...

PR #1601, opened 2015-11-19
adds support for custom typescript compiler options

dbldots:


lcd047:
This is not needed, you can use the standard variable `g:syntastic_typescript_tsc_args` instead.

mp-jgoetzinger:
@lcd047 oh. thanks for pointing out!

PR #1604, opened 2015-11-25
fix maven project which has set srcdir or testsrc dir

wsdjeg:
hi @lcd047 i have fix the error with maven project which has set src directory! 
if set src = src/
the out_put dir will be the project root director,here is an example,you can have a try
https://github.com/wsdjeg/android-GridViewWithHeaderAndFooter


lcd047:
Thank you.  Merged, with this patch on top: 10788ac

wsdjeg:
thank you !

PR #1605, opened 2015-11-27
exec mvn command only once

wsdjeg:
i find this script need to exec mvn command twice, it need a long time .only once will be ok,i have test this PR, works well


wsdjeg:
please waitting ,miss some func

PR #1606, opened 2015-11-27
Add dockerfile-lint checker

pixelastic:
This adds a syntaxchecker for `Dockerfile`. This uses [dockerfile-lint](https://github.com/projectatomic/dockerfile_lint) and expect the `dockerfile-lint` binary to be available (usually through a `npm -g` install).

By default, it will simply check for basic syntax correctness of the `Dockerfile`. One can pass a custom  `yml` rule file through the `syntastic_dockerfile_dockerfile_lint_post_args`, for example:

```vim
let b:syntastic_dockerfile_dockerfile_lint_post_args = '-r /path/to/rule/file.yml'
```

I've used `preprocess` for the json parsing part.

### DIFF
--- a/autoload/syntastic/preprocess.vim
+++ b/autoload/syntastic/preprocess.vim
@@ -59,6 +59,33 @@ function! syntastic#preprocess#cppcheck(errors) abort " {{{2
     return map(copy(a:errors), 'substitute(v:val, ''\v^\[[^]]+\]\zs( -\> \[[^]]+\])+\ze:'', "", "")')
 endfunction " }}}2
 
+function! syntastic#preprocess#dockerfile_lint(errors) abort " {{{2
+    let out = []
+    let json = s:_decode_JSON(join(a:errors, ''))
+    let data = json['error']['data'] + json['warn']['data'] + json['info']['data']
+
+    if type(json) == type({})
+        for e in data
+          let type = toupper(e['level'][0])
+          let type = (type ==# 'I') ? 'W' : type
+          let line = has_key(e, 'line') ? e['line'] : 1
+          let message = e['message']
+          if has_key(e, 'description') && e['description'] !=# 'None'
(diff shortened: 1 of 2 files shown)


lcd047:
Thank you.  Merged, with this patch on top: 73dd785.

pixelastic:
Thanks!

Could you tell me more about the `Style`/`%n`/`e['nr']` dance?

lcd047:
There isn't much to tell.  Normal languages have named regexps for parsing error messages, Vim has `errorformat`.  Besides being an obfuscated pain in the rear, `errorformat` is quite limited, there is very little information you can pass from an error message to the corresponding item in the resulting `quickfix` list.  Normally all you get is a filename, line and column numbers, an error type, and an error description.  If you need, say, a second column number, you're out of luck.  To get the job done, your options are to either (1) include the extra information in the error description, then parse it out from there after the `quickfix` has been built, or (2) abuse some unused %-format to pass the information.  Syntastic uses both techniques, and my patch above is an example of the latter one.

Normally `%n` in `errorformat` is used to parse error numbers.  It gives you something like `143 col 37 warning  302` in the `quickfix` window, and you can look up "warning 302" to find out more details about the error.  In the `quickfix` list itself it gives you a `nr`.  If the checker doesn't provide meaningful error numbers however, you can use `%n` to parse a number (say, a second column number).  But if you use it like this you need to delete `nr` from the `quickfix` items (after saving it elsewhere, of course), so that Vim doesn't show it to you as an error number.

In the patch above I use `nr` to signal that the corresponding items are information messages, and are to be marked as style warnings.  Setting `e['subtype']` to `Style` does the marking.  It's an unbelievably gross hack, but it gets the job done. _shrug_

pixelastic:
Haha, thanks for the details. I learn new vim quirks at every PR :)

PR #1608, opened 2015-11-28
correct pymode lint variable name in documentation

nixon:
The most recent versions of python-mode don't have a `pymode_lint_write` variable.  It appears to be called `pymode_lint_on_write` now.  I've updated the variable name in the syntastic doc so that it matches the accompanying text.

https://github.com/klen/python-mode/blob/master/doc/pymode.txt#l271


lcd047:
Merged, thank you.

PR #1610, opened 2015-11-28
Add slim-lint checker

justCxx:
This adds a syntaxchecker for *.slim files. This uses slim-lint
(https://github.com/sds/slim-lint) and expect the slim-lint binary to be
available (usually through a `gem install slim_lint`)


lcd047:
Thank you.  Merged, with this patch on top: 6d2e406.

Issue #1613, opened 2015-11-29
errors editing commit messages in git-for-windows


dgoldstein0:
I'm running syntastic, installed via pathogen, via the git-for-windows bash shell.  I have vim configured to be my default editor, and editing code seems to work fine but as soon as I try to `git commit` anything, I get this error:

> Cannot execute shell C:/Program Files/Git/usr/bin/bash
>
> syntastic: error: your shell C:/Program Files/Git/usr/bin/bash can't handle traditional UNIX syntax for redirections

and then several following errors.  It would appear, from looking at the code this error comes from (https://github.com/scrooloose/syntastic/blob/dc516ef19de71409aadd7d104c5f92312940c11b/plugin/syntastic.vim#L60).  Backtracking a little, it turns out that `syntastic#util#isRunningWindows()` does not recognize the git-for-windows bash environment as being windows.

lcd047:
Your setup is broken, Vim's `system()` can't use `C:/Program Files/Git/usr/bin/bash` as a shell.  Fix that, and syntastic will start working.

dgoldstein0:
My `system` command works just fine - to test this, I ran `:call system("echo 123 > 4.s")` then exited vim and found the 4.s file had been written with contents `123`.

There is definitely something fishy here with it trying to use ` C:/Program Files/Git/usr/bin/bash` instead of `/c/Program Files/Git/usr/bin/bash`.  Do you know where it would get this path? Afaik it can't come from my `$PATH` since that seems to always use the /c/ style.

dgoldstein0:
hm, I take that back.  For some reason, system is fried, but only in git commit messages (wtf?)

dgoldstein0:
So it seems my `SHELL` env variable has the wrong value when vim is invoked from `git commit`.  I'll work on tracking this down (and any pointers on why this might be would be appreciated), but this doesn't seem like a bug in syntastic but rather in my setup.

dgoldstein0:
it looks like `set shell=/usr/bin/bash` fixes this for me.  Dunno why git is messing up the `SHELL` variable though.

Chiel92:
For other people experiencing the same problem: I use the following snippet at the top of my vimrc to solve it:

```vim
if !has("gui_running")
    set shell=/usr/bin/bash
endif
```

This removes the error from the OP for me. However, after doing this, `:echo $SHELL` still says `C:/Program Files/Git/usr/bin/bash` when doing a `git commit`.

PR #1614, opened 2015-11-29
recognize git-for-windows as windows

dgoldstein0:
This seems to fix https://github.com/scrooloose/syntastic/issues/1613 for me


lcd047:
Please refrain from making changes without fully understanding what the code you're changing is supposed to do.  Your patch (1) removes the check without solving the problem, and (2) breaks other things.  Sorry, but no.

dgoldstein0:
I made this change, to my local version, and saw it silenced all the errors I was seeing (reported in #1613 ).  I knew there was a chance this might not be the right fix though

Issue #1493, opened 2015-07-29
gtk library not found


ScriptorNY:
Hello,

I am desperately trying to make the gtk library work with syntastic, but with no luck. Syntastic reports me errors about types, such as `uknown type name 'GtkApplication'`, or `implicit declaration of function`.

The source code I tested with :

```c
#include <gtk/gtk.h>

static void
activate (GtkApplication* app,
          gpointer        user_data)
{
  GtkWidget *window;

  window = gtk_application_window_new (app);
  gtk_window_set_title (GTK_WINDOW (window), "Window");
  gtk_window_set_default_size (GTK_WINDOW (window), 200, 200);
  gtk_widget_show_all (window);
}

int
main (int    argc,
      char **argv)
{
  GtkApplication *app;
  int status;

  app = gtk_application_new ("org.gtk.example", G_APPLICATION_FLAGS_NONE);
  g_signal_connect (app, "activate", G_CALLBACK (activate), NULL);
  status = g_application_run (G_APPLICATION (app), argc, argv);
  g_object_unref (app);

  return status;
}
```

Here is my syntastic config:
```vimL
" general syntastic config
let g:syntastic_check_on_open = 1
let g:syntastic_always_populate_loc_list = 1
let g:syntastic_auto_loc_list = 1

" c syntastic options
let g:syntastic_c_compiler = 'gcc'
let g:syntastic_c_config_file = '.config_c'
let g:syntastic_c_check_header = 0
let g:syntastic_c_no_include_search = 0
let g:syntastic_c_auto_refresh_includes = 1

```
Here is the content of my `.config_c` file : 
```
-Wall
-Werror
-Wextra
-I/usr/include/gtk-3.0
-I/usr/include/at-spi2-atk/2.0
-I/usr/include/at-spi-2.0
-I/usr/include/dbus-1.0
-I/usr/lib/dbus-1.0/include
-I/usr/include/gtk-3.0
-I/usr/include/gio-unix-2.0/
-I/usr/include/cairo
-I/usr/include/pango-1.0
-I/usr/include/atk-1.0
-I/usr/include/cairo
-I/usr/include/pixman-1
-I/usr/include/freetype2
-I/usr/include/libpng16
-I/usr/include/harfbuzz
-I/usr/include/freetype2
-I/usr/include/harfbuzz
-I/usr/include/libdrm
-I/usr/include/libpng16
-I/usr/include/gdk-pixbuf-2.0
-I/usr/include/libpng16
[...]

PR #1617, opened 2015-12-01
c.vim: add gtk+-3.0 to pkgconfig search

bhundven:
If gtk+-2.0 is not installed, check for gtk+-3.0.

Signed-off-by: Bryan Hundven <bryanhundven@gmail.com>


lcd047:
Not a good idea: if both GTK3 and GTK2 are installed, GTK2 includes will take precedence.  Moving GTK3 behind GTK2 is not a good idea either, because it's an incompatible change for a default.  Sorry but no; see [here](https://github.com/scrooloose/syntastic/issues/1493#issuecomment-125927220) for a workaround.

bhundven:
Thanks! I was kinda hoping I wouldn't have to "work around" an old version of gtk.

lcd047:
There are some design decisions in syntastic that no longer make sense today.  The same is true for many of the defaults.  However, they have been around since forever, and thus it would take some very strong reasons to change them.  The alternative is "WTF has changed today".  Maybe I'll get rid of all that junk in a single move at some point, probably just before a major release.

PR #1618, opened 2015-12-01
i have test with a lot of gradle project ,classpath detected works well!

wsdjeg:
TODO:
1. detect the classpath       DONE
2. detect the output dir       DONE
3. format
4.clear up code

### DIFF
--- a/syntax_checkers/java/classpath.gradle
+++ b/syntax_checkers/java/classpath.gradle
@@ -1,0 +1,40 @@
+task classpath << {
+
+  String finalFileContents = ""
+  HashSet<String> classpathFiles = new HashSet<String>()
+  for (proj in allprojects) {
+    
+    def exploded = proj.getBuildDir().absolutePath + File.separator + "intermediates" + File.separator + "exploded-aar"
+    def listFiles = new File(exploded)
+    if (listFiles.exists()) {
+      listFiles.eachFileRecurse(){ file ->
+        if (file.name.endsWith("classes.jar")){
+          classpathFiles += file
+        }
+      }
+    }
+
+    for (conf in proj.configurations) {
+        for (dependency in conf) {
+            if (dependency.name.endsWith("aar")){
+            } else {
+              classpathFiles += dependency
+            }
+        }
+    }
+    if (proj.hasProperty("android")){
+      classpathFiles += proj.android.bootClasspath
+    }
+
+    if (proj.hasProperty("sourceSets")) {
+    
+      for (srcSet in proj.sourceSets) {
+          for (dir in srcSet.java.srcDirs) {
+              classpathFiles += dir.absolutePath
+          }
+      }
+    }
+  }
+  def paths = classpathFiles.join(File.pathSeparator)
+  println "CLASSPATH:" + paths
+}
--- a/syntax_checkers/java/javac.vim
+++ b/syntax_checkers/java/javac.vim
@@ -17,6 +17,7 @@ let g:loaded_syntastic_java_javac_checker = 1
 let g:syntastic_java_javac_maven_pom_tags = ['build', 'properties']
 let g:syntastic_java_javac_maven_pom_properties = {}
 let s:has_maven = 0
+let s:has_gradle = 0
 
 let s:save_cpo = &cpo
 set cpo&vim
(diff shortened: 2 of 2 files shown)


wsdjeg:
@lcd047 can you check this PR,here is an test gradle project,https://github.com/wsdjeg/cube-sdk
before checking please use `./gradlew build` let gradle download the necessary jars like maven!

lcd047:
This needs some cleanup, but I think I cab deal wth that.

@troydm Any comments?

wsdjeg:
thank you ,i really want to do it,but i am afraid i will break the origin process

wsdjeg:
maybe `build/intermediates/classes/debug`should change to join('build','intermediates','classes','debug',sep)`

troydm:
@lcd047 @wsdjeg looks fine to me, as long as code that is related to Gradle support is handled separately the chance that you'll break something is minimal. I'm not too familiar with Gradle but if it handles dependency classpathes the same way as Maven does then adding this kind of support is handy, otherwise we are just addressing some specific Gradle project configuration that is not much usable outside of it's scope.

wsdjeg:
hi @lcd047 @troydm the func i have just added supports most gradle java project,i have test a lot of gradle android project,if there some issue,i would like to fix it!

lcd047:
Sorry, but [this](https://github.com/wsdjeg/syntastic/commit/55975beedfeff3dcee95ce3cf80b874d979c57ce#diff-3c44f2f1999397816acc71d07a9a4b03R457) is a show stopper.  Please consider maintaining it as an external checker.  It won't be merged in syntastic.

wsdjeg:
why,i am sure it can worked same sa maven

wsdjeg:
hi @lcd047
i have test it with a lot of java project ,it works well well,and now most java project use gradle instead of maven！

lcd047:
@wsdjeg The point is the vast majority of Java projects use Eclipse these days, rather than Vim.  The `javac` checker is already an order of magnitude more complex than any other checker in syntastic, and your patch adds even more complexity to it.  There are also some design decisions in your patch that I find questionable, regardless of how well it works in practice.  As I said, you can maintain it as an [external checkers][ext] if you absolutely need it.  External checkers are exactly the same as internal ones, the only difference is they are not distributed with syntastic.

[ext]: https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external

wsdjeg:
yes ，i agree with you,as you know javacomplete2 also can detect classpath,and the same thing should not be done twice！and i will create an external checker ,now i am reading external checker API

lcd047:
It's probably better to rename it to something other than `javac`, to avoid conflicts withnthe internal checkers.  That's all there is to it.

wsdjeg:
ok thank you

mschaaf:
@wsdjeg Hej I am very interested in your checker. Is it possible to download it in the external checker variant?

wsdjeg:
it is hard for me to write a external checker ,maybe you can use my fork for java checker
https://github.com/wsdjeg/syntastic

troydm:
Also check javac checker wiki page as it contains instructions on using gradle syntactic plugin that handles classpath for javac

wsdjeg:
the plugin do not contains the output dir,so it will miss the class of current project

troydm:
@wsdjeg well adding that capability to that plugin is possible, I think you should try communicating with it's author just to check if it's a trivial feature to implement or not

wsdjeg:
@troydm
to that plugin,we need change the code of the project,but not everyone use it,so i still think it is better add the gradle support into javac checker

lcd047:
> to use that plugin,we need change the code of the project

No you don't: all you need to do is to copy the existing checker, rename `javac` in the names of variables and functions to something else, and add your changes.

wsdjeg:
thanks ,i have tried to do it,but failed to load the checker created by myself

troydm:
> No you don't: all you need to do is to copy the existing checker, rename javac in the names of variables and functions to something else, and add your changes.
@lcd047 I don't think he is talking about checker itself, he means the https://github.com/Scuilion/gradle-syntastic-plugin I've mentioned

wsdjeg:
yes @troydm to use this gradle-syntastic-plugin ,i need to add groovy script into gradle project's build.gradle,but if the project is not mine,other people will not want me add this script.

PR #1619, opened 2015-12-02
Syntax Checker: Add rudimentary support for .travis.yml files with syntax=travisci

joelfrederico:
Travis CI has a configuration file .travis.yml. (See https://github.com/travis-ci/travis.rb#lint) Assuming the user has set the syntax to travisci, we can use the 'travis lint' command to check the file.

However, this checker is rather awful. It isn't designed to be parsed and this issue has been closed with wontfix: https://github.com/travis-ci/travis.rb/issues/203.

But I think this is better than nothing.

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -88,6 +88,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'tex':           ['lacheck', 'chktex'],
         \ 'texinfo':       ['makeinfo'],
         \ 'text':          [],
+        \ 'travisci':      ['travis'],
         \ 'twig':          ['twiglint'],
         \ 'typescript':    ['tsc'],
         \ 'vala':          ['valac'],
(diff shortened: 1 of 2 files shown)


lcd047:
As far as I can tell "awful" is somewhat of an understatement.  However, I do disagree with you that adding crap to syntastic is better than nothing.  Sorry about that.  You can maintain it as an [external checker][ext] if you must, I have a few of those myself.

On a side note, I'm not sure using the API as suggested in the thread you reference would be much better than parsing the output of `travis` and / or `travis-lint`.  The response to `lint` doesn't seem to include relevant information about line numbers either. _shrug_

[ext]: https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external

joelfrederico:
Sad story, but thanks for getting back to me fast. I think using an online API for any syntax checker is a terrible idea. What if the internet's out? Or the machine's firewalled? What about speed?

I'll go with the external checker then. Even just knowing there is an error saves me lots of time and mistakes, so it's useful for me.

PR #1627, opened 2015-12-04
Add R Lintr support

jimhester:
I have actually had R script support working for over a year (see https://github.com/jimhester/lintr/blob/master/inst/syntastic/lintr.vim) however never contributed the file back upstream (sorry!).

I attempted to add support for Rmd files as well (lintr handles them in the same way as R files), however I must not have something quite right because the Rmd linter is not properly registered by syntactic.

I am happy to make changes if you have time to review the code. Thanks for maintaining a wonderful package!

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -71,6 +71,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'python':        ['python', 'flake8', 'pylint'],
         \ 'qml':           ['qmllint'],
         \ 'r':             [],
+        \ 'rmd':           [],
         \ 'racket':        ['racket'],
         \ 'rnc':           ['rnv'],
         \ 'rst':           ['rst2pseudoxml'],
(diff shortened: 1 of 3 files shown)


lcd047:
Thank you for the clean patch.

> I attempted to add support for Rmd files as well (lintr handles them in the same way as R files), however I must not have something quite right because the Rmd linter is not properly registered by syntactic.

This was actually a limitation in the registry code, `exec` was not inherited automatically from the old checker, so you had to set it explicitly.  I removed this limitation in e60555d.  It's also no longer needed to call
```vim
runtime! syntax_checkers/<filetype>/*.vim
```
to bootstrap redirected checkers (but doing it anyway is harmless, so this doesn't add any API incompatibility).

lcd047:
Please add descriptions to the checker options in the wiki for [r/lintr](https://github.com/scrooloose/syntastic/wiki/R%3A---lintr) and [rmd/lintr](https://github.com/scrooloose/syntastic/wiki/R---Markdown%3A---lintr).

PR #1641, opened 2015-12-19
Add --force-exclusion flag to rubocop call

RST-J:
This fixes what I thought to be an [issue](https://github.com/bbatsov/rubocop/issues/2492#issuecomment-164375887) of Rubocop itself.

Summary: Rubocop allows to configure the behaviour of its checkers by a config file (`.rubocop.yml`). E.g. you can disable particular checkers for certain files or paths (this surprisingly worked already before my patch) or you can disable all checkers for certain files or paths which did not work without the `--force-exclusion` flag.

### DIFF
--- a/syntax_checkers/ruby/rubocop.vim
+++ b/syntax_checkers/ruby/rubocop.vim
@@ -26,7 +26,7 @@ function! SyntaxCheckers_ruby_rubocop_IsAvailable() dict
 endfunction
 
 function! SyntaxCheckers_ruby_rubocop_GetLocList() dict
-    let makeprg = self.makeprgBuild({ 'args_after': '--format emacs' })
+    let makeprg = self.makeprgBuild({ 'args_after': '--format emacs --force-exclusion' })
 
     let errorformat = '%f:%l:%c: %t: %m'
 


lcd047:
This is not needed: you can achieve the exact same effect with `syntastic_ruby_rubocop_args`:
```vim
let g:syntastic_ruby_rubocop_args = '--force-exclusion'
```
It's also wrong to add it `args_after`, since `args_after` options can't be overridden by users.  Sorry, but no.

albb0920:
Quoting from the issue from rubocop

> There's an option called --force-exclusion, which editor plugins should use. The problem is that when you give a single file name to rubocop, then Exclude parameters don't come into play. Unless you run with --force-exclusion.

I see this has been rejected twice (also #1973),
since setting that option is the expected way to use rubocop when integrating with an editor,
is there any chance to have  this option set by default? (eg. by changing the default syntastic_ruby_rubocop_args)

lcd047:
@albb0920 Sorry, but no.  Don't get me wrong, I have nothing against `--force-exclusion`.  But defaults should never change in incompatible ways, if at all possible.  Syntastic isn't aiming to work out of the box.

PR #1646, opened 2015-12-21
Don't check preview window

kghost:


lcd047:

kghost:
syntastic check syntax and open location-list for preview window, which is unnecessary.

lcd047:
Syntastic needs a file of a known filetype to even consider running checks.  If your preview window actually corresponds to a file on disk, then whatever set it that way is badly broken.  Also, your patch is wrong.  But whatever, it's an easy fix, please see 281b294.

PR #1654, opened 2016-01-03
JSCS checker: Removing warning when receiving empty json response.

scottis:
Removing the warning that is printed when an empty json response is received from jscs.


lcd047:
This is the wrong fix.  Please see 3280220 instead.

PR #1662, opened 2016-01-09
Added stylelint syntax checker for scss file types from css stylelint syntax checker.

dylansm:
It's nearly identical to css checker, but uses scss file type and passes an additional flag: -s "scss" (syntax: "scss").


dylansm:
Looks like this might be better refactored as indicated here: https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#duplicates

lcd047:
Yes, the way described in the guide is preferable.  Commit 1f4846f.

PR #1663, opened 2016-01-13
Added rubber-info checker for tex

slollo:


lcd047:
`rubber-info` is mainly a log parser.  This is not a problem by itself, but the checker doesn't do anything to make sure the logs are up to date after you make changes to the sources.  This is not apropriate for syntastic.  There are also other problems, f.i. the checker looks for `rubber` instead of `rubber-info` for activation, and lines such as the following are ignored by your `errorformat`:
```
vuldmexp.tex:170-172: Overfull \hbox (2.01886pt too wide) (page 3)
```
Sorry, but no.

Issue #1664, opened 2016-01-13
D syntax checker import fails with DSFML Lib


afaur:
Since the import checker is hard checking `source` folder, but not `src` folders `syntastic` reports import errors when your code is correct.  When you correct this by being more specific it will pass the syntastic checker, however it will break the build.

Syntactic Version: 
```
Syntastic version: 3.6.0-93 (Vim 704, Darwin)
```
Vim Version:
```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Aug 27 2015 15:03:01)
MacOS X (unix) version
Included patches: 1-769
```

The example library that is trying to be imported is DSFML:
```
-> % pwd
~/.dub/packages/dsfml-2.1.0 
-> % ls
CONTRIBUTING.md     README.md           dub.json            libdsfml_audio.a    libdsfml_network.a  libdsfml_window.a   unittest
License.txt         build.d             libdsfml.a          libdsfml_graphics.a libdsfml_system.a   src
-> % cd src
-> % ls
dsfml 
-> % cd dsfml
-> % ls
audio    graphics network  system   window
-> % cd graphics
-> % ls
blendmode.d      drawable.d       package.d        renderstates.d   shader.d         texture.d        vertexarray.d
circleshape.d    font.d           primitivetype.d  rendertarget.d   shape.d          transform.d      view.d
color.d          glyph.d          rect.d           rendertexture.d  sprite.d         transformable.d
convexshape.d    image.d          rectangleshape.d renderwindow.d   text.d           vertex.d
-> % cat package.d 
/*
DSFML - The Simple and Fast Multimedia Library for D

Copyright (c) 2013 - 2015 Jeremy DeHaan (dehaan.jeremiah@gmail.com)

This software is provided 'as-is', without any express or implied warranty.
In no event will the authors be held liable for any damages arising from the use of this software.

Permission is granted to anyone to use this software for any purpose, including commercial applications,
and to alter it and redistribute it freely, subject to the following restrictions:

1. The origin of this software must not be misrepresented; you must not claim that you wrote the original software.
If you use this software in a product, an acknowledgment in the product documentation would be appreciated but is not required.

2. Altered source versions must be plainly marked as such, and must not be misrepresented as being the original software.

3. This notice may not be removed or altered from any source distribution
*/

///2D graphics module: sprites, text, shapes, etc.
module dsfml.graphics;

public
{
	import dsfml.window;

	import dsfml.graphics.blendmode;
[...]

PR #1665, opened 2016-01-13
#1664 - D syntax checker import fails with DSFML Lib 

afaur:
- Add `src` directory to `g:syntastic_d_include_dirs`.
- Fix for issue https://github.com/scrooloose/syntastic/issues/1664


lcd047:
Patches that change default values are accepted only if you can make a very strong case for them.  In this particular situation, adding includes from source packages has negative side effects.  Sorry, but no.

PR #1666, opened 2016-01-14
Added configuration for jscs if it's available in the project folder

nemisj:
Explicit specifying config file path

### DIFF
--- a/syntax_checkers/javascript/jscs.vim
+++ b/syntax_checkers/javascript/jscs.vim
@@ -29,7 +29,8 @@ function! SyntaxCheckers_javascript_jscs_IsAvailable() dict
 endfunction
 
 function! SyntaxCheckers_javascript_jscs_GetLocList() dict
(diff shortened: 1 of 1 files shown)


lcd047:
This is not a good idea.  It breaks backward compatibility for no reason, it prevents the use of a config file with a name different from `.jscsrc`, and it can be achieved with standard Vim mechanisms without patching syntastic.  It's also an attempt to use syntastic to work around a usability problem in JSCS.  Sorry, but no.

nemisj:
@lcd047, can you please tell me how I can achieve this with standard Vim mechanisms so that i can have my jscs@2.8.0 working with .jscsrc in our project folder by using :w ?

lcd047:
One way to do it:
```vim
autocmd BufNewFile,BufReadPre *.js let b:syntastic_javascript_jscs_args =
    \ '--config ' . findfile('.jscsrc', escape(expand('%:p:h'), ' ') . ';')
```

nemisj:
thanks :) will useit

PR #1674, opened 2016-01-25
Added a flag for javac_config_file to use git root to find the config…

glentakahashi:
Having to open vim from the root of the git repo in order to get the functionality of .syntastic_javac_config is pretty annoying, so I added a flag to have syntastic check to see if the file exists at the root of the git repo regardless of where you open vim.

### DIFF
--- a/syntax_checkers/java/javac.vim
+++ b/syntax_checkers/java/javac.vim
@@ -55,8 +55,16 @@ if !exists('g:syntastic_java_javac_config_file_enabled')
     let g:syntastic_java_javac_config_file_enabled = 0
 endif
 
+if !exists('g:syntastic_java_javac_config_file_use_git_root')
+    let g:syntastic_java_javac_config_file_use_git_root = 0
+endif
+
 if !exists('g:syntastic_java_javac_config_file')
-    let g:syntastic_java_javac_config_file = '.syntastic_javac_config'
+    if g:syntastic_java_javac_config_file_use_git_root
+        let g:syntastic_java_javac_config_file = system('echo -n `git rev-parse --show-toplevel`') . '/.syntastic_javac_config'
+    else
+        let g:syntastic_java_javac_config_file = '.syntastic_javac_config'
+    endif
 endif
 
 if !exists('g:syntastic_java_javac_custom_classpath_command')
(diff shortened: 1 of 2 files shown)


lcd047:
About the `javac` checker:

* you can set `syntastic_java_javac_config_file` with an `autocmd`, without changing syntastic; this is consistent with Vim's approach everywhere else;
* you don't need to run `git` to find the root of the project; searching upwards for `.syntastic_javac_config` (or for `.git`) would be enough;
* the checker is already an order of magnitude more complicated  than any other one; it should be simplified, not made even more complex;
* the Java environment is not friendly to command-line applications; if you need more features than syntastic offers you should consider switching to an IDE.

As for adding support for `tsx`:

* please don't mix unrelated patches in PRs;
* changes to defaults are not accepted without very strong reasons; in this case you can achieve the same by setting `syntastic_typescript_tsc_args`, without changing syntastic;
* learning how to take advantage of `autocmd` would save you a lot of time.

Sorry, but no.

glentakahashi:
Thanks for the feedback @lcd047 !
I forgot that pushing back to develop would add new changes to the PR so sorry for that.

Good points all around, I'll look into using `autocmd`. I guess I should have started an issue thread before doing a random PR. (Still getting into the hang of contributing to public repos)

glentakahashi:
@lcd047 I can't find the flag syntastic_typescript_tsc_args or something that would parse the args properly. It seems as though they're hardcoded. Would that be a valid PR to submit?

lcd047:
See `:help syntastic-config-makeprg`.  But you probably want something like this instead:
```vim
autocmd BufNewFile,BufReadPre *.js  let g:syntastic_java_javac_config_file =
    \ findfile('.syntastic_javac_config', escape(expand('%:p:h'), ' '))
```

glentakahashi:
I ended up using this which I found from `:help findfile()`

```
" Search upwards for .syntastic_javac_config
autocmd BufNewFile,BufReadPre *.java let g:syntastic_java_javac_config_file = findfile('.syntastic_javac_config', '.;')
```

lcd047:
The difference between your `autocmd` and the one I wrote above is that mine starts looking for a `.syntastic_javac_config` in the base directory of the current file, while yours starts from the current directory.

PR #1675, opened 2016-01-26
Add YAML checker: yamllint

adrienverge:
yamllint [1] is a linter for YAML files. It does not only check for
syntax validity, but for common cosmetic conventions such as lines
length, trailing spaces, indentation, etc.

Also set yamllint as the default YAML linter, since other checkers
(js-yaml and yamlxs) do not do linting but solely detect syntax errors.

1: [http://yamllint.readthedocs.org/](http://yamllint.readthedocs.org/)

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -100,7 +100,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'xslt':          ['xmllint'],
         \ 'xquery':        ['basex'],
         \ 'yacc':          ['bison'],
-        \ 'yaml':          ['jsyaml'],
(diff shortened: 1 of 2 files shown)


lcd047:
The license of your checker (GPL v3) is more restrictive than the license of the project (WTFPL v2).

adrienverge:
> The license of your checker (GPL v3) is more restrictive than the license of the project (WTFPL v2).

License changed!

> No changes to the defaults please.

I restored `jsyaml` as the "default default". However, `jsyaml` is a parser that can only detect syntax errors, whereas `yamllint` is a real linting tool (syntax + configurable rules). Can you ack this is what you want?

lcd047:
Thank you.  Merged, with a few changes.  Please see 8c4dadc and cc440f1.

Issue #960, opened 2014-01-25
Add support for finding Dub packages in D source code.


TyOverby:
Right now when I edit D code that uses an external library that is managed by the Dub package manager, I always get an error on my imports (and nowhere else) because dmd doesn't know about the existence of these libraries.

This could be solved by running dub (if it exists) with "dub describe" to get a list of the paths to the dependencies.  Then dmd could be run with those paths known.

PR #1691, opened 2016-02-11
Determine exact import paths by running dub for D linting.

w0rp:
The all but official package manager for the D programming language is DUB. DUB allows for project configuration with requirements for packages in the DUB package registry, and storing packages in the user's home directory, etc. The DUB executable knows where all of the modules for a project are,so it can build them.

Some time ago, I added an option to DUB so you can ask it for the _exact_ import paths it will use to build a project, as absolute paths, no matter where they happen to be. (`dub describe --import-paths`) This command has been in the stable versions of DUB since September. It's not always enough to simply set up a glob match for source directories in `~/.dub/packages/*/source`, etc., as package maintainers can name the source directories with any name they desire. (And _they do_...)

This pull request searches for the project root when editing D source files, where DUB is installed, and uses DUB to determine exactly where all of the modules live. In this manner, it should improve linting by always being able to locate all of the modules which are being imported.

### DIFF
--- a/syntax_checkers/d/dmd.vim
+++ b/syntax_checkers/d/dmd.vim
@@ -36,12 +36,47 @@ function! SyntaxCheckers_d_dmd_IsAvailable() dict
 endfunction
 
 function! SyntaxCheckers_d_dmd_GetLocList() dict
-    if !exists('g:syntastic_d_include_dirs')
-        let g:syntastic_d_include_dirs = filter(glob($HOME . '/.dub/packages/*', 1, 1), 'isdirectory(v:val)')
-        call map(g:syntastic_d_include_dirs, 'isdirectory(v:val . "/source") ? v:val . "/source" : v:val')
-        call add(g:syntastic_d_include_dirs, './source')
+    let dubPaths = []
+
+    if executable('dub')
+        " If dub is installed, start searching for the project root,
+        " from the current source directory.
+        let path = expand('%:p:h')
+        let dubCommand = 'dub describe --import-paths'
+
+        while 1
+            if len(globpath(path, 'dub.json')) > 0 || len(globpath(path, 'dub.sdl')) > 0 || len(globpath(path, 'package.json')) > 0
+                " We hit a directory with a dub package file.
+                " Run the command to find the paths.
+                let dubPaths = split(system('cd ' . shellescape(path) . ' && ' . dubCommand), '\n')
+
+                if v:shell_error
+                    " The dub command failed, so clear the captured output.
+                    " Otherwise, we will capture some garbage.
+                    let dubPaths = []
+                endif
+
+                break
+            endif
+
+            let nextPath = fnamemodify(path, ':h')
+
+            if path == nextPath
+                " We just checked a root directory, so stop here.
+                break
+            endif
+
+            " Go up one directory.
+            let path = nextPath
+        endwhile
     endif
 
+    if len(dubPaths) > 0
+        " We found the exact import paths through DUB, so use those.
+        return dubPaths
+    endif
+
+    " Fall back on using DMD to find import paths.
     return syntastic#c#GetLocList('d', 'dmd', {
         \ 'errorformat':
         \     '%-G%f:%s:,%f(%l): %m,' .


lcd047:
I presume this is a messed up rebase: your patch doesn't even make sense, let alone work as intended. :)

Regardless, I'm afraid nothing revolutionary happened since #960, my conclusion there still stands.  Running `dub` at every check is too slow, `syntastic_d_include_dirs` should be set by a shell script that runs `dub` as needed, and caches the include directories per project.  A SQLite backend might be appropriate too, for extra geek credits.  It _has_ to be written as an external script, since you don't want to reinvent `make` in Vim.  Sorry, but no. _shrug_

w0rp:
This was absolutely not a "messed up rebase." It did precisely what I intended. It changed the `SyntaxCheckers_d_dmd_GetLocList()` function to ask `dub` for the locations of all of the import paths.

If you say it doesn't "work as intended," did you even run it with the most recent version of `dub`? It works perfectly for me. If you did you know know it is not, in fact, "too slow." `dub` can tell you exactly which import paths are being used very quickly.

The problem with setting `syntastic_d_include_dirs` is that it **doesn't work**. You end up missing out directories from the import paths, or you end up including modules which aren't in your project at all, or you end up having to manually reconfigure your global variable (which applies to every project) each time you add a new module.

Sorry, but I shall maintain a fork of this project from this point forward which actually works.

lcd047:
> This was absolutely not a "messed up rebase." It did precisely what I intended.

The `*_GetLocList()` function is supposed to return (surprise!) a loclist.  Yours returns a list of paths.

> If you say it doesn't "work as intended," did you even run it with the most recent version of `dub`?

I didn't need to.  Did you?

w0rp:
>The *_GetLocList() function is supposed to return (surprise!) a loclist. Yours returns a list of paths.

Right, now that actually makes sense. I'll move the production of the import paths to the part where it handles that, and then pass those on to the compiler.

lcd047:
> I'll move the production of the import paths to the part where it handles that, and then pass those on to the compiler.

Setting `syntastic_d_include_dirs` is the API way to do that.

w0rp:
I have amend the commit now to change the list returned from `s:_get_include_dirs`, and now that works. Thanks for letting me know. I'll keep it on my fork, and I can maybe put it in a repo for an external checker with `syntax_checkers/d/dub.vim` or something.

lcd047:
> I have amend the commit now to change the list returned from `s:_get_include_dirs`

That breaks all other checkers that happen to use `syntastic#c#GetLocList()`.  Just saying.

w0rp:
It checks if the `filetype` is `d` and if `dub` is executable.

lcd047:
... and then there's a proper way to do this, without any code changes:
```vim
function Dubious(base)
    let where = escape(a:base, ' ') . ';'

    let old_suffixesadd = &suffixesadd
    let dirs = syntastic#util#unique(map(filter(
        \   findfile('dub.json', where, -1) +
        \   findfile('dub.sdl', where, -1) +
        \   findfile('package.json', where, -1),
        \ 'filereadable(v:val)'), 'fnamemodify(v:val, ":h")'))
    let &suffixesadd = old_suffixesadd

    let includes = []
    for dir in dirs
        try
            execute 'silent lcd ' . fnameescape(dir)
            call extend(includes, split(system('dub describe --import-paths', "\n")))
            silent lcd -
        catch /\m^Vim\%((\a\+)\)\=:E472/
            " evil directory is evil
        endtry
    endfor

    return syntastic#util#unique(includes)
endfunction

autocmd BufRead,BufNewFile *.d,*.di let g:syntastic_d_include_dirs = Dubious(expand('<amatch>:h', 1))
```
(not tested).

w0rp:
@lcd047 I have added another commit which does what you pasted above, and adds a new option called `g:syntastic_d_automatic_dub_include_dirs` for turning it on in a .vimrc file, etc. Could something like this be included in the main repo? It would be nice if others could use it easily.

See here: https://github.com/w0rp/syntastic/commit/5a1db797e5bb8bfda46cd2107d6e8758750609b3

lcd047:
> I have added another commit which does what you pasted above

The code above is supposed to be added to your `vimrc`.  Adding an `autocmd` to the checker itself is basically a time bomb.

> Could something like this be included in the main repo?

Not in the current form, but I'll eventually do it if I get positive feedback from people using the checker.

@TyOverby, @benekastah, @jakob-lundberg, @yamadapc, @cheesington, @afaur: any comments on this?

belka-ew:
Yes this function seems to work. I only had to remove the second argument  from syntastic#util#system call, otherwise it claims about too many arguments.
dub describe --import-paths is available only in the current stable, 0.9.24: https://github.com/D-Programming-Language/dub/blob/master/CHANGELOG.md

PR #1694, opened 2016-02-12
Coverity desktop analysis checker

Letme:
Parsing the Coverity cov-run-desktop commands on all files which were
changed according to scm from last time. It would be better, to remove
the scm flag and just pass the file to buffer, but still need to figure
out how.
Syntax marks CID's (defects) as errors, how it came to them as
information and few special misra violations as warnings.

### DIFF
--- a/syntax_checkers/c/coverity.vim
+++ b/syntax_checkers/c/coverity.vim
@@ -1,0 +1,55 @@
+"============================================================================
+"File:        coverity.vim
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Crt Mori <crt at the-mori dot com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
+"             See http://sam.zoy.org/wtfpl/COPYING for more details.
+"
+"============================================================================
+
+if exists('g:loaded_syntastic_c_coverity_checker')
+    finish
+endif
+let g:loaded_syntastic_c_coverity_checker = 1
+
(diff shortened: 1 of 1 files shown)


lcd047:
This seems awkward at best, and is pretty specialized.  It also has a few technical problems (f.i. syntastic can only handle errors and warnings, not info messages).  Perhaps you should maintain it as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).

Letme:
I must say I am surprised, as info messages displayed nicely and were also
bound to lines. There was no familiar red or yellow indicator (could be
blue), but it still worked nicely. Since this steps are information how to
reach the error message I would not promote them to warnings.
How about if we add info messages in some way?

lcd047:
> I must say I am surprised, as info messages displayed nicely and were also bound to lines.

There's a _lot_ more to error levels in syntastic than looking nice in a single test on your machine.  Syntastic never supported info messages.  I don't know why; I suppose syntastic's author didn't find them useful back when he started the project.  Could they be added now?  Not really.  The changes to the core would be nontrivial, and 50+ affected checkers would have to be updated and tested.  Well, working PRs are welcome. :)

> How about if we add info messages in some way?

Syntastic has style messages, and they can be used to distinguish between more message types than just errors and warnings.

However, the entire workflow is awkward in other ways.  I'm not ready to support this inside syntastic; I'm afraid it will have to be an external checker.  There isn't any loss of functionality in that: external checkers are functionally identical to internal ones for all practical purposes.

Letme:
Since I have created external checker as suggested I have put it into a new directory [on GitHub](https://github.com/Letme/syntastic_coverity) . Just in case someone would want to use it.

lcd047:
Please read the docs.  You need a certain directory structure.  Syntastic won't pick it up otherwise.

Issue #1383, opened 2015-04-12
Using Syntastic Vim plugin with the Linux kernel


vladimiroltean:
Hi,

This is not an issue, but I've been trying to properly configure Syntastic for the Linux kernel, and there's not one online tutorial that I could find about this, so I figured this might be of help.

Vladimir

https://github.com/vladimiroltean/blog/wiki

lcd047:
> `:%s/\ /\r/g`

Nit pick: this is better written as `:%s/ /\r/g`.

> * by default, Syntastic treats include paths as relative to the configuration file. Obviously, we need to change those paths so that they are absolute (prefix them with `$(KDIR)` as described above, in our case `/usr/linux/so2/`)

That would be something like this:
```vim
:%s!\m^-I\zs\ze[^/]!/usr/linux/so2/!
```

> * for some reason, Syntastic's interpretation of `-D` macro predefinitions is different than gcc. It can be seen in the last 3 lines of the above text that KBUILD_BASENAME and KBUILD_MODNAME depend upon KBUILD_STR, but Syntastic doesn't know how to resolve that.

You don't understand.  Syntastic config files does exactly two things, and both are trivial: (1) change `-Irelative/path` to `-I/absolute/path`, and (2) escape lines to survive shell expansion when passed to the compiler.  If you write `-D"KBUILD_STR(s)=#s"`, the result of escaping is `'-D"KBUILD_STR(s)=#s"'`, which means after shell expansion gcc actually gets to see the argument `-D"KBUILD_STR(s)=#s"`.  What you want it to see instead is `-DKBUILD_STR(s)=#s`, and that's what you have to write in the config file.  Of course, the same applies to the other `-D`s.

> `let g:syntastic_c_config_file = '.syntastic_c.conf'`

Since you just wrote a bunch of options to a file named `.syntastic_c_config`, this is confusing. :)

> To check what we have done, write something to a C file, then `:w` to save it.

It's probably far more informative to enable debugging and look at the arguments syntastic passes to gcc: set `g:syntastic_debug` to 1, run the checker, run `:mes`, and look for `makeprg`.  This way, you'll probably find out you also need:
```vim
let g:syntastic_c_no_default_include_dirs = 1
let g:syntastic_c_compiler_options = ''
```

> If all goes well, you should now have Syntastic working with the Linux kernel.

You might want to rephrase this.

vladimiroltean:
[...]

PR #1695, opened 2016-02-12
Linux makefile syntax checker with fixed arguments

Letme:
I saw the Issue scrooloose/syntastic#1383 and decided that it might be
useful to contribute my simple makefile for linux kernel development.

Changing a comment allows me to change the arguments passed to the Linux
Kernel makefile, path in cwd allows me to switch between different
forks. All the rest was copied from standard makefile syntax and
displays errors and warnings nicely.

### DIFF
--- a/syntax_checkers/c/makekernel.vim
+++ b/syntax_checkers/c/makekernel.vim
@@ -1,0 +1,64 @@
+"============================================================================
+"File:        makekernel.vim
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Crt Mori <crt at the-mori dot com>
+"License:     This program is free software. It comes without any warranty,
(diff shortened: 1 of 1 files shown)


lcd047:
If afraid this is not needed: you can achieve the exact same thing without code changes, by setting `syntastic_c_make_args` and `syntastic_c_make_fname` in your vimrc:
```vim
let g:syntastic_c_make_args = '-j4'
let g:syntastic_c_make_fname = 'ARCH=arm CROSS_COMPILE=arm-none-eabi- am335x_boneblack42_defconfig zImage dtbs'
```
You can also use an `autocmd` to set these only for kernel sources.

In general, patches that add checkers for the only reason to change defaults are not accepted.  Sorry, but no.

PR #1696, opened 2016-02-14
add markdownlint checker

davidosomething:
requires https://github.com/igorshubovych/markdownlint-cli npm package

configurable using a json file:

      let b:syntastic_markdown_markdownlint_args = '--config markdownlint.json'

different from mdl in that it runs using node and has more configuration options for rules (but otherwise same rules). Also the errorfmt is different by one space.

### DIFF
--- a/syntax_checkers/markdown/markdownlint.vim
+++ b/syntax_checkers/markdown/markdownlint.vim
@@ -1,0 +1,42 @@
+"============================================================================
+"File:        markdownlint.vim
+"Description: Syntax checking plugin for syntastic.vim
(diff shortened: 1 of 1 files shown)


davidosomething:
never mind, ppl can just drop this into ~/.vim/syntax_checkers/markdown/ if they want it

lcd047:
Or update to 0bdf149.

davidosomething:
@lcd047 you gotta get rid of the `--warnings` arg too -- cli flags are different

lcd047:
Yes, that's the kind of information that is provided in the [wiki](https://github.com/scrooloose/syntastic/wiki/Markdown%3A---mdl).

PR #1697, opened 2016-02-15
Remove erroneous line from FAQ

jaredmanning:
The line in question "__4.7. Q. I run a checker and the location list is not updated...__" preceded a question that was almost, but more detailed, with the same number as it.


lcd047:
Nope, this is intentional: both problems have the same cause, and the answer apply to both questions.

jaredmanning:
Ah, sorry about that. Thanks for the clarification.

PR #1698, opened 2016-02-15
Checker rst/rst2pseudoxml: errors without a line #

maxwell-k:
Stop a type of error failing silently; for example if a document
includes:

    `Link text`__

but misses the corresponding:

    .. __: http://www.example.org

then without this patch, syntastic shows no errors. `rst2pseudoxml`
produces output like:

    file.txt:2: (ERROR/3) Anonymous hyperlink mismatch: 1 references but 0 targets.
    See "backrefs" attribute for IDs.

This patch adds a pattern to 'errorformat' to catch this type of error
message that doesn't include a line number.

### DIFF
--- a/syntax_checkers/rst/rst2pseudoxml.vim
+++ b/syntax_checkers/rst/rst2pseudoxml.vim
@@ -31,7 +31,12 @@ function! SyntaxCheckers_rst_rst2pseudoxml_GetLocList() dict
     let errorformat =
         \ '%f:%l: (%tNFO/1) %m,'.
         \ '%f:%l: (%tARNING/2) %m,'.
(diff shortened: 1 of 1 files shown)


lcd047:
It isn't entirely clear whether this is the right solution or not.  Loclist items without line numbers are mostly useless, since they don't allow you to move to the corresponding point in the sources.  Syntastic doesn't handle them very well, either.  They should be reserved for things like fatal conditions, when you want to know about the problem but you can only do something about it outside Vim.  In this particular case I'd say this would be better addressed by fixing it in `rst2pseudoxml`.  But if you want error lines without line numbers, you can have them.  Commit c3882ef.

maxwell-k:
Thanks!

PR #1708, opened 2016-02-26
NASM syntax-checker : -f elf to args_after to be able to override them

GNUletik:
This allows to override the ```-f``` option in ```.vimrc```.
For example : ```let g:syntastic_nasm_nasm_args = "-f elf64"```

### DIFF
--- a/syntax_checkers/nasm/nasm.vim
+++ b/syntax_checkers/nasm/nasm.vim
@@ -20,8 +20,8 @@ set cpo&vim
 
 function! SyntaxCheckers_nasm_nasm_GetLocList() dict
     let makeprg = self.makeprgBuild({
-        \	'args_before': ' -f elf' ,
-        \ 	'args_after': ' -X gnu' .
+        \ 	'args_after': ' -X gnu' ,
+        \	'post_args': ' -f elf' .
         \       ' -I ' . syntastic#util#shescape(expand('%:p:h', 1) . syntastic#util#Slash()) .
         \       ' ' . syntastic#c#NullOutput() })
 
--- a/syntax_checkers/nasm/nasm.vim
+++ b/syntax_checkers/nasm/nasm.vim
@@ -20,7 +20,8 @@ set cpo&vim
 
 function! SyntaxCheckers_nasm_nasm_GetLocList() dict
     let makeprg = self.makeprgBuild({
-        \ 'args_after': '-X gnu -f elf' .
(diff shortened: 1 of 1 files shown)


lcd047:
> This allows to override the `-f` option in `.vimrc`.

It doesn't.  Use `args` or (better, in this particular case) `post_args` instead of `args_before` and `args_after` for arguments you want to override.  Enable debugging to see what syntastic is doing.

GNUletik:
Sorry, I should  have explain the point of this.
I want to be able to replace the ```-f``` option of NASM in my .vimrc.

The normal way is to add this line in .vimrc : 
```
let g:syntastic_nasm_nasm_args = "-f elf64"
```

The problem is that it doesn't work because the option is overridden by the args_after option in the NASM plugin.

So, to fix this we can put this parameter in the args_before like I did here.

lcd047:
> Sorry, I should have explain the point of this.

You did.

> So, to fix this we can put this parameter in the args_before like I did here.

No.  The proper way to do this is to move `-f elf` to an option that can be overridden (rather than duplicated with a new value) by the user.  `args_before` and `args_after` can't be overridden, so the candidates are `args` and `post_args`.  Changing `args` breaks backwards compatibility for people that set `g:syntastic_nasm_nasm_args` in their vimrc.  This leaves you with `args_after`.  If you move `-f elf` there it can still be overridden by `g:syntastic_nasm_nasm_post_args`.

lcd047:
Alternatively, you can just set
```vim
let g:syntastic_nasm_nasm_post_args = '-f elf64'
```
without any patch. _shrug_

GNUletik:
Okay, I see the point !

I didn't know the user was able to override post_args. Should I do another pull request to add it into the README.md ?

I thought the traditional args shouldn't be used anymore in the syntax checker.

So, you're right, as we can use post_args as user it simpler to use this way.
I moved the ```-f elf``` option into post_args to be cleaner anyway.

Thanks !

lcd047:
> I didn't know the user was able to override post_args. Should I do another pull request to add it into the README.md ?

It's documented, but sadly the information is spread across several files.  From a user's percpective: `:h 'syntastic_<filetype>_<checker>_<option>'`; from a checker writer's perspective: see the [guide](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#makeprgbuildoptions).  Basically `args` and `post_args` are user-definable, while `*_before` and `*_after` are checker internals.

Issue #1688, opened 2016-02-04
HLint not working


afonsojoao96:
Hi there, i've installed `hlint` in my Ubuntu 14.04 machine (with `apt-get install hlint`, not cabal), however when i edit a haskell file it does not perform syntax check.

Running `:SyntasticInfo` i get:
```
Syntastic version: 3.7.0-81 (Vim 704, Linux)
Info for filetype: haskell
Global mode: active
Filetype haskell is active
The current file will be checked automatically
Available checker: hlint
Currently enabled checker: hlint
```
So i assume that Syntastic knows that i've hlint installed and working.
`:SyntasticCheck` does nothing with a file containing obvious errors.

Hlint is working good when i run it through my terminal.
Vim version: 7.4

Any thoughts about that?

lcd047:
> when i edit a haskell file it does not perform syntax check.

Please open your test file, set `g:syntastic_debug` to 3, run the checker, then run `:mes` and post the output.

afonsojoao96:
Wow it is now working, i simply ran:
`let g:syntastic_debug=3` followed by
`:mes`
and now it highlights the errors!
Thanks dude

PR #1711, opened 2016-02-28
Support new HLint version

bhipple:
In the [latest version of HLint](http://neilmitchell.blogspot.com/2016/02/new-hlint-version-and-api-features.html), linting warnings are now classified as "Suggestion".

Tested with Stack on a file with a handful of warnings and suggestions.  This could be the underlying cause of #1688, which was posted a few days after the new HLint was released (though only if the user had a cabal or stack installed version ahead of the dpkg version in his PATH).


lcd047:
Merged, thank you.

PR #1704, opened 2016-02-20
Support for the renaming of Jade to Pug

acidjazz:
Due to this [unfortunate incident](https://github.com/pugjs/jade/issues/2184), [these fine people](https://www.jadeworld.com/) have trademarked a [rock](https://en.wikipedia.org/wiki/Jade). Forcing the rename of the Jade language to [Pug](https://github.com/pugjs), which I think is a compliment to the company that trademarks minerals since the CIO [resembles the breed](https://www.jadeworld.com/media-center/cio-john-ascroft-traditional-businesses-going-digital/)

PR #1712, opened 2016-02-28
Add pydocstyle checker for Python

rmk135:
Due the fact that Python `pep257` tool was renamed, I propose to add appropriate syntax checker for the new name with leaving old name "as is" for leaving backward compatibility.

Links:
- https://github.com/PyCQA/pydocstyle/issues/172
- https://github.com/PyCQA/pydocstyle/pull/173


lcd047:
Right: encouraged by the overwhelming success of the transition from Python 2 to Python 3, Python's powers that be are now targeting the smaller prey. :)

Anyway, this is similar to #1704.  Commit c6d6c79.

PR #1716, opened 2016-03-02
Update to match latest version of js-yaml

zanona:
Apparently a few things have changed with the latest version and he `compact` flag no longer outputs line and column numbers.
I have removed the additional flag and adjusted the error message prefix to match with https://github.com/nodeca/js-yaml (3.5.3).
Hopefully that  can be helpful.

### DIFF
--- a/syntax_checkers/yaml/jsyaml.vim
+++ b/syntax_checkers/yaml/jsyaml.vim
@@ -19,15 +19,10 @@ let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_yaml_jsyaml_GetLocList() dict
-    if !exists('s:js_yaml_new')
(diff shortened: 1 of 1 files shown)


lcd047:
What problem does this solve, aside from breaking compatibility with older versions of js-yaml?

zanona:
I wasn't being able to use the linter with the current version. Apparently the error flag is not prefixed with `YAMLException` instead `JS-YAML`?

lcd047:
Please post a test file that works on the command line and breaks in syntastic.

zanona:
Sorry about that, I have tested it again, but the issue might have been coming from elsewhere because the current version works without a problem now. Sorry for taking your time with this confusion. I will make sure to test more efficiently next time.

PR #1719, opened 2016-03-07
Attempt to detect CM files in SML/NJ checker

jez:
When editing SML/NJ files inside projects that use Compilation Manager for build
management, Syntastic displays errors incorrectly because it makes no attempt to
look for these `*.cm` files.

This pull request lets the smlnj checker find the CM file and appropriately
configure the makeprg when these one of these files is present.

Let me know if this is a feature you'd like or not like to include in Syntastic.
I'm open to suggestions. I wrote a bit of documentation in the file itself; I'm
not sure if we'd need to update any documentation elsewhere if this were to be
merged.

### DIFF
--- a/syntax_checkers/sml/smlnj.vim
+++ b/syntax_checkers/sml/smlnj.vim
@@ -18,8 +18,56 @@ let g:loaded_syntastic_sml_smlnj_checker = 1
 let s:save_cpo = &cpo
 set cpo&vim
 
+" If the user is using a CM file to manage their program, we'd like to detect
+" this and set the flags for makeprgBuild appropriately.
+"
+" This function basically automates this process which formerly required user
+" intervention:
+"
+"   - Figure out whether the project is using a CM file
+"   - Figure out the name of the CM file for this project
+"   - Edit the vimrc, using the appropriate name:
+"         let g:syntastic_sml_smlnj_args = "-m sources.cm"
+"         let g:syntastic_sml_smlnj_fname = ""
+"   - Re-launch Vim
+"
+" Of course none of these are that hard, but it's tedious to have to edit the
+" vimrc every time you start working in a different folder on a new project.
+
+function! s:DetectCMFile()
+    " Start searching at the current folder
+    let curdir = fnamemodify("%", ":p:h")
+
+    while 1
+        " Check if there are any cm files in the current folder
+        let cmfiles = split(globpath(curdir, "*.cm"))
+        if len(cmfiles) > 0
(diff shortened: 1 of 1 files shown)


lcd047:
Actually, you can achieve the same thing without changes to the code.  Something like this:
```vim
function! s:DetectCM(fname) abort
    let cm = syntastic#util#findGlobInParent('*.cm', fnamemodify(a:fname, ':p:h'))
    if cm !=# ''
        let buf = bufnr(fnameescape(a:fname))
        call setbufvar(buf, 'syntastic_sml_smlnj_args', '-m ' . syntastic#util#shescape(cm))
        call setbufvar(buf, 'syntastic_sml_smlnj_fname', '')
    endif
endfunction

autocmd FileType sml        call s:DetectCM(expand('<amatch>', 1))
```
Or perhaps put the relevant piece of code in a `ftplugin`.  Either way, it's the kind of manipulation that doesn't make sense for everybody, and thus it's better kept out of syntastic.  Sorry, but no.

jez:
Thanks for the tip! I had no idea that `findGlobInParent` was already a function; that certainly makes it simpler.

For those who end up finding this issue from a search: I've made this into a plugin (with a few other goodies for SML in general): [jez/vim-better-sml].

@lcd047 Out of curiosity, is the existence of `syntastic#util` documented anywhere? I checked over the wiki documentation on writing a checker but didn't see any mention there.

Thanks again for the snappy response!

lcd047:
> I've made this into a plugin

I sent you a PR.

If you add initialisation to your plugin you'll also need to make sure syntastic is loaded before the rest of the code gets to run.  This is not a problem for a `ftplugin` though, at least not until somebody tries to get smart and thinks to use the new `:packadd` command to delay loading syntastic until it's actually needed.

> Out of curiosity, is the existence of `syntastic#util` documented anywhere?

It depends on what you mean by that.  `:h autoload` will explain the basics.  Beyond that, well, nobody took up the task of documenting the `util` functions, and the API is still subject to change; some people still seem to have high hopes that syntastic will be able to do async checks some day.

jez:
> If you add initialisation to your plugin you'll also need to make sure syntastic is loaded before the rest of the code gets to run.

I thought about that; I'll probably deal with it if it ever becomes an issue.

> It depends on what you mean by that.

I was more referring to which functions were available in `util`. If the API is still subject to change, it's understandable that it's undocumented. I guess I'll just keep this in mind if my plugin breaks after an update :stuck_out_tongue:

PR #1725, opened 2016-03-15
Pass --dev flag to jruby for faster startup

kbarrette:
Pass the `--dev` flag to jruby to (slightly) improve startup time, as documented here: https://github.com/jruby/jruby/wiki/Improving-startup-time

Caveats: this only works with newer versions of jruby, not sure what the best way to do a version check is.


lcd047:
> Caveats: this only works with newer versions of jruby, not sure what the best way to do a version check is.

A version check would cold run `jruby --version`, which slow down everything to a crawl.  Consequently, the best way to handle this is to leave it to the user to add `--dev` to `g:syntastic_ruby_jruby_args` or `g:syntastic_ruby_jruby_post_args`.  Perhaps add a note to that effect to the [wiki](https://github.com/scrooloose/syntastic/wiki/Ruby%3A---jruby). _shrug_

PR #1728, opened 2016-03-16
Change scss-lint executable to scss_lint

glittershark:
see https://github.com/brigade/scss-lint/commit/c1e311b495cfd189f63aa8cc821ca036b47a9fe9 - this is now the proper executable per Gem name standards


lcd047:
I read the link you pointed me to, and I reached a different conclusion.  Syntastic cares about executables rather than gem names, and as of 5b2dbfd the executable is still named [`scss-lint`](https://github.com/brigade/scss-lint/blob/e6d20f537d021445da9583cbe1ec29bcded5ec32/bin/scss-lint).  When / if that changes, please post a new PR with proper version checks and fallbacks for backwards compatibility.  Until then, sorry but no.

PR #1740, opened 2016-04-09
D checker: Add src to default source paths

belka-ew:
src and source are both default source paths in DUB.


lcd047:
That used to be the default in the past, and people had trouble with it.  Also, you can achieve the same thing without code changes, by setting `g:syntastic_d_include_dirs` in your vimrc.  Patches to the defaults are considered only when they solve serious problems.  Sorry, but no.

Please see [here](https://github.com/scrooloose/syntastic/pull/1691#issuecomment-182875063) for a better solution.  I'll make a patch to that effect if (1) you can confirm this works, and (2) you tell me which was the first (released) version of `dub` that could do `dub describe --import-paths`.

belka-ew:
No problem. I wasn't even sure it is the right way to do that. I added the function from the linked issue to my .vimrc. Will report next week if it works.

lcd047:
Please see cc6b92a.

PR #1741, opened 2016-04-10
Add typescript tsc support add external compiler option

elct9620:
Hi, I using TypeScript to write ReactJS.
But the `tsc` alwasy warning me need to use `--jsx` compiler options to use with `tsx`.
So I add some patch to current TypeScript syntax checker to allow me add compiler options.

### DIFF
--- a/syntax_checkers/typescript/tsc.vim
+++ b/syntax_checkers/typescript/tsc.vim
@@ -14,6 +14,10 @@ if !exists('g:syntastic_typescript_tsc_sort')
     let g:syntastic_typescript_tsc_sort = 1
 endif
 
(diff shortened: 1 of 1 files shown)


lcd047:
This is not needed, you can achieve the same effect by setting `g:syntastic_typescript_tsc_args` in your vimrc.  Sorry, but no.

elct9620:
Ok, I look it is working for me.
Is there have any way to add some document for this?
This setting seems cannot found in source code or exists document.

lcd047:
> Is there have any way to add some document for this?

`:h 'syntastic_<filetype>_<checker>_<option>'`

PR #1742, opened 2016-04-11
Fix grammatical error in README.markdown

mgeraci:
Fix grammatical error: "none ... is" => "none ... are"


lcd047:
Merged, thank you.

PR #1748, opened 2016-04-14
Absolute filepath for error and warnings.

JulienPivard:
On save file, if the file is in project, open new buffer with bad filepath so the buffer open is empty.
Before change:
```shell
gcc -c -x ada -gnats src/principale.adb
principale.adb:32:26: missing ";"
```

quickfix try to open principale.adb but file is in src/ not in current directory.

After change:
```shell
gcc -c -x ada -gnats -gnatef src/principale.adb 
src/principale.adb:32:26: missing ";"
```

quickfix open file src/principale.adb.


lcd047:
Merged, thank you.

PR #1751, opened 2016-04-20
Kotlin gradle syntax checker

tomm:


lcd047:
1. Please add a license.
2. The `IsAvailable()` function is not needed.
3. You don't need to escape spaces in `errorformat`.
4. You don't need an `exec` in `CreateAndRegisterChecker()` if it's the same as `name`.

lcd047:
On second thought, it's probably better for you to maintain this [outside syntastic](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).  I have only a vague idea what Kotlin is, and I can't really test that your checker works at all.  Sorry, but no.

PR #1753, opened 2016-04-23
checker govet: use makeprgBuild

petertseng:
The current [wiki page][1] states that the govet checker doesn't use the
makeprgBuild function. I couldn't find a documented reason (in wiki,
commit messages, or in #616) that indicates why it must not use
makeprgBuild, so this commit makes it so. It's probably because the
checker was added in April 2013, before makeprgBuild was added in
November 2013.

The specific reason this is desirable is that otherwise `go vet` is
invoked without any extra arguments, meaning that it runs on the entire
set of files in the current directory. This causes it to output warnings
for files that are not the current file; the current file may be clean
but I may still see warnings for other files.

By using makeprgBuild, in addition to respecting
`g:syntastic_go_govet_<option>`, it would also only run on the current
file.

[1]: https://github.com/scrooloose/syntastic/wiki/Go:---govet

### DIFF
--- a/syntax_checkers/go/govet.vim
+++ b/syntax_checkers/go/govet.vim
@@ -23,7 +23,7 @@ function! SyntaxCheckers_go_govet_IsAvailable() dict
 endfunction
 
 function! SyntaxCheckers_go_govet_GetLocList() dict
-    let makeprg = self.getExecEscaped() . ' vet'
+    let makeprg = self.makeprgBuild({ 'args_before': 'vet' })
 
     let errorformat =
         \ '%Evet: %.%\+: %f:%l:%c: %m,' .


petertseng:
I admit maybe "I'm holding it wrong" and the go vet checker is *intended* to be run on all the files in the current directory, in which case I will gladly hear alternatives on how to suppress warnings that are not for the current file.

syntax_checkers/go/govet.vim
```diff
@@ -23,7 +23,7 @@ function! SyntaxCheckers_go_govet_IsAvailable() dict
 endfunction
 
 function! SyntaxCheckers_go_govet_GetLocList() dict
-    let makeprg = self.getExecEscaped() . ' vet'
+    let makeprg = self.makeprgBuild({ 'args_after': 'vet' })
```
petertseng: could possibly just be `args` not `args_after`? Couldn't figure out, actually! Seems to work with just `args` though...

petertseng: I looked at `g:SyntasticChecker._getOpt` briefly, seems it's safe to just use `args`. Using `args`, unless you tell me otherwise.

petertseng:
Small example of the situation I am trying to deal with: I might have two files:

a.go

```go
package main

import "fmt"

func main() {
	fmt.Printf("Hello, %s\n")
}
```

b.go

```go
package main

func aaaaaa() {
}
```

If we run `go vet` without arguments, it says:

```
a.go:6: missing argument for Printf("%s"): format reads arg 1, have only 0 args
```

Thus if I am editing `b.go`, my status bar displays `[Syntax: line:6 (1)]` Even though `b.go` has no line 6. (There is *correctly* no marker in the gutter though, the marker correctly only shows if I edit `a.go`). I would like that `[Syntax: line:6 (1)]` to not show up if I am looking at `b.go`, so I propose this PR.

syntax_checkers/go/govet.vim
petertseng: and now I read https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide , looks like `args` can be changed by user and `args_before` and `args_after` can't. So I have decided to use `args_before`.

lcd047:
> I admit maybe "I'm holding it wrong" and the go vet checker is *intended* to be run on all the files in the current directory,

This is a line of thought worth exploring.

> in which case I will gladly hear alternatives on how to suppress warnings that are not for the current file.

This is probably a bad idea in most situations, but if you actually _want_ it, one way to do it might be something like this:
```vim
autocmd FileType go
    \ let b:syntastic_go_govet_quiet_messages =
    \     { '!file:p': '\V' . escape(expand('<afile>:p'), '\') }
```
Believe it or not, this is actually documented (see `:h 'syntastic_<filetype>_<checker>_quiet_messages'`).

> Thus if I am editing `b.go`, my status bar displays `[Syntax: line:6 (1)]`

This is correct, if you run `:Errors` you'll see the error in the loclist.

> Even though `b.go` has no line 6.

Syntastic doesn't promise to show you something that makes sense regardless of how you configure it.  If you want syntastic to give you more discriminating information on the status line you might consider investing a minute in reading the relevant part of the manual: `:h syntastic-statusline-flag` and `:h 'syntastic_stl_format'`.

> thus with this PR

That breaks the far more common case of checking modules.  Sorry, but no.

petertseng:
Thanks for the tips and pointers to manual, I have found the sections you point to and will read and decide how to configure for me.

Reading about go vet I have learned that it is indeed useful to provide it the entire package not single files.

It could be possible to both 1) preserve the entire package checking behavior and 2) use makeprgBuild which will allow the standard `g:syntastic_go_govet_<option>` to work if using `self.makeprgBuild({ 'args_before': 'vet', 'fname': '' })` then. Let me know if that seems suitable as well.

Issue #1692, opened 2016-02-11
eslint path problems


acthp:
As noted in another issue, global eslint will not run the eslint version installed in a project. The recommended way to run the correct version is via npm, like 'npm run lint', after creating a lint script directive in package.json, like {'lint': 'eslint'}.

However, setting javascript_eslint_exec to 'npm run lint' in syntastic results in syntastic not finding the executable. Setting it to 'npm' and javascript_eslint_args to 'run lint --' results in command line

    npm -f compat run lint -- foo.js

which is wrong. It needs to be

    npm run lint -- -f compat foo.js

The 'args_before' call botches the order.

We need better control of the command line to make it work.

lcd047:
`javascript_eslint_exe`

acthp:
Thank you!

    let g:syntastic_javascript_eslint_exe = 'npm run eslint --'

This works.

gurpreetatwal:
If you don't want to add a custom task to your package scripts the following setting will always run the locally installed eslint

```
let g:syntastic_javascript_eslint_exe = '$(npm bin)/eslint
```

stuartsan:
Note that may help someone else: for me the approaches described by @acthp and @gurpreetatwal, which are aimed at using a project's local version of `eslint`, did not work until I _also_ had `eslint` installed globally (`npm install -g eslint`).

lcd047:
@stuartsan Just point `*_exec` (not `*_exe`) to a valid binary:
```vim
let g:syntastic_javascript_eslint_exec = '/bin/ls'
```

stuartsan:
Ahh gtk, thanks @lcd047 !

kogakure:
I wrote a solution to use the local node_module script or fallback to the global installation, when not available:

```vim
let g:syntastic_javascript_eslint_exe = '[ -f $(npm bin)/eslint ] && $(npm bin)/eslint || eslint'
```

lcd047:
@kogakure Alternatively, write a wrapper script and point `g:syntastic_javascript_eslint_exec` to it.

ezmiller:
I have tried a couple of these examples for getting syntastic to use the local eslint but to no avail. The above: 

`let g:syntastic_javascript_eslint_exe = '[ -f $(npm bin)/eslint ] && $(npm bin)/eslint || eslint'

Led to output like this:

```
syntastic: 7.221055: g:syntastic_version = '3.8.0-14 (Vim 704, Neovim, Darwin)'
[...]

PR #1756, opened 2016-04-30
Use local eslint when possible

tran2:
Some projects/node modules have the following structure
- project
  - node_modules
    - eslint
    - eslint plugins
  - src
    - javascript
  - package.json
  - .eslintrc (specify what eslint plugins to use)

This patch changes the eslint exec path to the local node_module's eslint
`path-to-project/node_modules/eslint/bin/eslint.js`
in place of `eslint`
When local path is not found, we'll use eslint as default.

The reason for change:
- global eslint looks for eslint plugins globally rather than from inside the node_modules and will throw plugins missing errors

### DIFF
--- a/syntax_checkers/javascript/eslint.vim
+++ b/syntax_checkers/javascript/eslint.vim
@@ -22,6 +22,16 @@ if !exists('g:syntastic_javascript_eslint_generic')
     let g:syntastic_javascript_eslint_generic = 0
 endif
 
+" Use local eslint when possible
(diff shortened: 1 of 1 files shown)


lcd047:
1. There is no need to change the code, the same can be accomplished with an `autocmd`.
2. Your patch is awkward, and it breaks when switching to a different project from within Vim.
3. There is a better way, see #1692.

Sorry, but no.

PR #1763, opened 2016-05-05
add turtle and trig check via rapper

seebi:
turtle and trig are RDF syntaxes defined by the W3C
https://www.w3.org/TR/turtle/
https://www.w3.org/TR/trig/

rapper is an RDF parsing and serializing utility
http://librdf.org/raptor/rapper.html


lcd047:
Thank you.  Merged, with this patch on top: 71208eb.

PR #1766, opened 2016-05-08
Choose dialect based on vim syntax

keith:
Shellcheck is smart enough to check the shebang in a given file to
determine which dialect to use. Unfortunately this doesn't work for
files without shebangs, even if it might be apparent what dialect should
be used, such as "bashrc" or "foo.bash". Luckily `filetype.vim` defines
specific vars based on which shell dialect is being used based on a huge
list of conditions. With this change we take those into account for all
the types shellcheck supports, otherwise we fallback to letting it try
and decide.

### DIFF
--- a/syntax_checkers/sh/shellcheck.vim
+++ b/syntax_checkers/sh/shellcheck.vim
@@ -11,8 +11,21 @@ let g:loaded_syntastic_sh_shellcheck_checker = 1
 let s:save_cpo = &cpo
 set cpo&vim
 
+function! s:GetDialectArgument()
+    if exists('b:is_bash') && b:is_bash
+        return '-s bash'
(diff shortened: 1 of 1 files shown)


lcd047:
This could be achieved without changes to the code.  In this particular case however the changes seem useful for everybody, so I'll include them.  Still, on a quick check:

* you're forcing a shell dialect even when there is a shebang line; I wouldn't quite trust Vim's syntax highlighting to be _always_ better than Shellcheck's hand-crafed detection;
* you look for `b:` variables but ignore any corresponding `g:` variables;
* you don't allow the user to override the dialect detected this way.

So please try this instead: 822cb14.

keith:
Thanks!

PR #1769, opened 2016-05-12
Add turtle check via ttl

selliera:
turtle in an RDF syntax defined by the W3C
https://www.w3.org/TR/turtle/

ttl is a nodejs turtle validation tool:
https://www.npmjs.com/package/turtle-validator

As a node application, it is very easy to install on windows, as well as any other platform.


lcd047:
Thank you.  Merged, with this patch on top: c2c6a07.

PR #1770, opened 2016-05-13
Enable the ttl checker by default

selliera:
It would be nice to have the ttl checker enabled by default, if available.


lcd047:
Sorry, patches that changes defaults are not accepted.  People are supposed to read the manual and set the `syntastic_*_checkers` to fit their use case, rather than rely on defaults.

selliera:
Ok, fine by me.
However, having this one enabled by default would be nice, since it is so easy to install.

lcd047:
Which is exactly why patches that changes defaults are not accepted: everybody wants their favourite checker to be autodetected, then opening a file slows down to a crawl, and everybody complains that syntastic is slow and invasive. _shrug_

PR #1772, opened 2016-05-16
Adds two Clean syntax checkers (cpm and clm)

camilstaps:
This PR adds a [Clean](http://clean.cs.ru.nl/Clean) syntax checker.

Unfortunately, Clean project management is a little complicated. This has complications for the syntax checker, because there is no external checker that will *just* check the syntax of the current file. `clm` will break when it cannot find imports; `cpm` requires a project file.

The syntax checker now tries the following to build the current module, say `module.icl`:

 1. If `module.prj` exists, run `cpm project module.prj build`
 2. If any other `*.prj` exists, say `project.prj`, run `cpm project project.prj build`
 3. Otherwise, run `clm -c module`

The catch is that when a project file exists in the current directory, but that project doesn't use the currently loaded file, the syntax of that file will not be checked.

Thanks to @timjs for the errorformat (https://github.com/timjs/vim-clean/blob/timjs-syntax/compiler/cpm.vim).


lcd047:
> This has complications for the syntax checker, because there is no external checker that will *just* check the syntax of the current file.

Syntastic doesn't _require_ checking a single file.

> `clm` will break when it cannot find imports;

Which is probably why you should at least consider the `syntastic#c#GetLocList()` framework.

> `cpm` requires a project file.

Syntastic provides functions `syntastic#util#findFileInParent()` and `syntastic#util#findGlobInParent()` to help you with that.

>  3. Otherwise, run `clm -c module`

But your checker doesn't do that, does it.

There are many problems with your patch:

* it assumes Vim is run in a certain directory; that's a show stopper;
* it makes automatic choices without giving the user any chance to override them; that's another show stopper: a feature that can't be turned off is a bug;
* it doesn't do proper escaping for shell meta characters (ok, you probably didn't know that);
* it falls into some common traps, such that not taking into account `wildignore` when running `expand()` (ok, normal people probably don't know that);
* the logic is slightly flawed.

camilstaps:
Thanks for the comments! Discussing this morning with other Clean users we got to some other ideas to change this checker, so it's good that it isn't merged yet :)

Most of your comments are clear and I will implement them. 

> > Otherwise, run clm -c module

> But your checker doesn't do that, does it.

What do you mean by this? If I'm not mistaken, the else block:

                    \ 'exe': 'clm',
                    \ 'args': '-c',
                    \ 'fname': module })

does this.

camilstaps:
This commit splits up the checker in two:

* clm, which will always simply try `clm -c module`
* cpm, which will do `cpm project module.prj build` if module.prj exists and otherwise `cpm make` in the first parent directory with a *.prj file

It also fixes the wildignore trap, and changes the working directory before checking.

Could you help me a bit more?

> it doesn't do proper escaping for shell meta characters

* Where concretely? In the makeprg? Is it sufficient to use `shellescape` on the arguments of `makeprgBuild`?

> Which is probably why you should at least consider the `syntastic#c#GetLocList()` framework.

* Sorry, but I fail to understand what this function does. Is there somewhere more documentation on this function, besides the comments?

The problem that I was referring to is the following: in order to be able to check the syntax, clm needs to be able to find the imported files. If clm cannot find these files, it will simply abort checking (because if some definitions are missing it can't do much without guessing, which would be difficult to implement). So, the syntax checker only works in the case that only standard libraries (or modules from the same directory) are used.

* Is the setup with a separate errorformat.vim OK like this? I wasn't sure how to solve this (sharing variables between two checkers).

lcd047:
> changes the working directory before checking.

Nope, the `cwd` option won't work the way you think.  The paths of the files in the loclist will be wrong.

> Could you help me a bit more?

Not if it takes 25 more exchanges back and forth to get the basics running.  Sorry about that.  Have you considered maintaining this as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external)?

> Is it sufficient to use `shellescape` on the arguments of `makeprgBuild`?

The value of `fname` in `makeprgBuild` is supposed to be escaped (if you choose to override it, that is).

Also, `args` and `post_args` are supposed to be options that can be overridden by the user.  This doesn't seem to be your intention here.

> Sorry, but I fail to understand what this function does.

It's a framework intended for compilers.  Well, if you don't understand it, it's probably better to leave it alone.

> Is there somewhere more documentation on this function, besides the comments?

None, sorry.  This is a one-man show these days, and I'm doing for free it in my spare time.  Many, many things are left undocumented.

> If clm cannot find these files, it will simply abort checking (because if some definitions are missing it can't do much without guessing, which would be difficult to implement).

Then perhaps you should be able to set some kind of include path?  If you can't do that, that would be a design problem with `clm` (or whatever).

BTW, if your checker relies on features present only in some versions of `clm` you need to add a version check.

> Is the setup with a separate errorformat.vim OK like this?

Not really.  Please don't try to optimize operations (i.e. editing `errorformat` in two places) that should have once every few years.

camilstaps:
Many thanks for your help. I didn't realise you were doing all this by yourself, I understand that you cannot spend time educating everyone. I hope we can finish this soon now and don't need to make it an external checker.

The latest commit puts into work all your comments:

 - escaped arguments to `makeprgBuild`
 - `args` and `post_args` can now be overridden
 - for the `clm` checker, it is now possible to provide a list of libraries `b:syntastic_clean_clm_libraries` that should be added to the include path. It is also possible by editing `args` directly, but this seems slightly easier because it automatically adds the necessary `-I` before each path addition
 - moved the `errorformat` back to the separate checkers

except perhaps this one:

> Nope, the `cwd` option won't work the way you think. The paths of the files in the loclist will be wrong.

I fixed an error where the wrong module name, and now all seems to work, although I didn't change anything about the way `cwd` is used. It is possible to compile a module in another directory than the current, also with imports from that directory. The paths in the loclist then appear as relative to the working directory (so when I start with `vim path/to/test.icl`, the loclist will have entries of `path/to/test.icl`). This is the correct way, isn't it?

`clm` has not had a stable release since the release of Clean 2.4 in 2011. I didn't check if the features I'm using were already available before 2011, but don't think anyone is still using anything below Clean 2.4. `cpm` is from the development release and the features I'm using have been implemented over a year ago (I didn't check further). Anyway, if people are on a development release, they should update regularly. So, I'd suggest version checks aren't necessary.

lcd047:
> escaped arguments to `makeprgBuild`

Syntastic provides function `syntastic#util#shescape()` instead of the Vim's `shellescape()`.

> for the `clm` checker, it is now possible to provide a list of libraries

Please don't add checker-specific variables for no reason.  There are a few of those that are kept around for backwards compatibility, and maintaining them is problematic at best.

> I fixed an error where the wrong module name, and now all seems to work, although I didn't change anything about the way `cwd` is used.

The way you use `cwd` is guaranteed __not__ to work.  It goes something like this:

* your current file is in `dir1`
* `cwd` changes current directory to `dir2` (presumed different from `dir1`)
* syntastic gets a loclist with filenames relative to `dir2`
* syntastic actually needs files relative to `dir1`.

> `clm` has not had a stable release since the release of Clean 2.4 in 2011. I didn't check if the features I'm using were already available before 2011, but don't think anyone is still using anything below Clean 2.4.

Reason enough to add a version check for 2.4 or later. _shrug_ Syntastic has `syntastic#util#versionIsAtLeast()` and `getVersion()` for that purpose.

camilstaps:
I now finally

camilstaps:
Thanks, I now finally see what you're getting at concerning `cwd`. The problem seems to be only for `cpm` in the case that it runs `cpm make` in a higher directory (in other cases `dir1` and `dir2` are equal).

However, debugging this I found a much more important problem: filenames in the output of clm/cpm are only the basename, not the path relative to some point or an absolute path. Hence there is no way to trace this back... this is a problem that needs to be solved in those programs before a syntax checker will really work, I'm afraid.

I'm going to go to the Clean team with this - until then there is not much to do here.

Sorry for wasting your time for this. Hopefully this can be made to work soon.

The latest commit does put into work your other comments (`syntastic#util#shescape()`, removed code for adding libraries to command line, add version check).

lcd047:
Ok, then please post a clean PR if / when `cln` and `cpm` are fixed.

In the mean time, please refrain from "fixing" documentation you [don't fully understand](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide/_compare/352d8cd3e6ae35def333e13719f44357ba5f7998...a6ab66fc5eb2623a41d59357286bc99ecc587102), thank you.

camilstaps:
Ah, apologies, I thought the function had moved.

PR #1776, opened 2016-05-20
Adds scss load-path for Rails projects

jadencarver:
I want to start a discussion around this.  When working on a Rails project, you will get a lot of 'could not find import' errors because the stylesheet asset path in Rails is app/assets/stylesheets - Is there already a way, or can we create a way to specify these paths?

### DIFF
--- a/syntax_checkers/sass/sass.vim
+++ b/syntax_checkers/sass/sass.vim
@@ -44,7 +44,7 @@ function! SyntaxCheckers_sass_sass_GetLocList() dict
     endif
 
     let makeprg = self.makeprgBuild({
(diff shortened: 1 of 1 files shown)


lcd047:
The same can be accomplished without any changes to the code, by setting `syntastic_sass_sass_args` in your vimrc:
```vim
let g:syntastic_sass_sass_args = '-I app/assets/stylesheets'
```
> I want to start a discussion around this.

Feel free to discuss it.  The root of the problems is `sass` returning inconsistent relative paths in error messages, and this has been known for a long time. _shrug_

jadencarver:
Thank you, that's exactly what I was looking for. It didn't pop out at me looking at the code, I'm still familiarizing myself with it.

That works for me, I only mentioned a discussion because I didn't want anybody to actually merge this.

Thank you.

PR #1777, opened 2016-05-21
Add gtk+-3.0 to the list of packages to look for.

alekulyn:
Syntastic couldn't find my local gtk3 pc file, so I just added it in the list of packages.


lcd047:
This has been discussed in the past, please see #1493 and #1617.  Sorry, but no.

PR #1778, opened 2016-05-21
Added g:syntastic_typescript_tsc_target option.

mkulke:
Annoyingly `tsc` ignores `tsconfig.json` files (in which a target can be set) when input source files are specified: http://www.typescriptlang.org/docs/handbook/tsconfig-json.html

To use es6 features like async/await without syntax warnings (`TS1311: Async functions are only available when targeting ECMAScript 6 and higher.`) I added g:syntastic_typescript_tsc_target.

@jb55


lcd047:
You can achieve the same effect without code changes, by setting `syntastic_typescript_tsc_args` in your vimrc.  Sorry, but no.

PR #1784, opened 2016-06-05
Fix ansible-lint errorformat

omarkhan:
See commit https://github.com/willthames/ansible-lint/commit/52a9e0b5f3f0df4d2a9a092ecf6935def7a3e5cf: an `E` has been added to the ansible-lint output string.


lcd047:
Right, but there's no need to break support for older versions along with the fix.  Commit 0ec2a5c.

PR #1785, opened 2016-06-05
Checker nvcc: use makeprgBuild and add config file support

nmoehrle:
The nvcc checker is currently not using the makeprgBuild function and thus does not allow users to set compiler arguments within vim configs. There is also no support for config files.
This pull request adds support for both.

I am not an experienced vimscript user and help as well as a code review would be most welcome, although most of this is copied from other checkers...

### DIFF
--- a/syntax_checkers/cuda/nvcc.vim
+++ b/syntax_checkers/cuda/nvcc.vim
@@ -22,13 +22,14 @@ function! SyntaxCheckers_cuda_nvcc_GetLocList() dict
         \ 'args_before': '--cuda -O0 -I . ' .
         \ syntastic#c#ReadConfig(g:syntastic_cuda_config_file) .
         \ ' -Xcompiler -fsyntax-only',
-        \ 'tail_after': syntastic#c#NullOutput()}
+        \ 'tail': syntastic#c#NullOutput()}
 
     if index(['h', 'hpp', 'cuh'], expand('%:e', 1), 0, 1) >= 0
         if syntastic#util#var('cuda_check_header', 0)
-            let buildoptions.exe_before = 'touch /tmp/.syntastic_dummy.cu ;'
-            let buildoptions.fname = '/tmp/.syntastic_dummy.cu'
+            let buildoptions.exe_before = 'echo > .syntastic_dummy.cu ;'
+            let buildoptions.fname = '.syntastic_dummy.cu'
             let buildoptions.args_after = '-include ' . syntastic#util#shexpand('%')
+            let buildoptions.tail_after = '; rm .syntastic_dummy.cu'
         else
             return []
         endif
--- a/syntax_checkers/cuda/nvcc.vim
+++ b/syntax_checkers/cuda/nvcc.vim
@@ -1,8 +1,8 @@
 "============================================================================
 "File:        cuda.vim
 "Description: Syntax checking plugin for syntastic
-"Author:      Hannes Schulz <schulz at ais dot uni-bonn dot de>
-"
+"Authors:     Hannes Schulz <schulz at ais dot uni-bonn dot de>
+"             Nils Moehrle <nils at kuenstle-moehrle dot de>
 "============================================================================
 
 if exists('g:loaded_syntastic_cuda_nvcc_checker')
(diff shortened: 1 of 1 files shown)


lcd047:
This would have been better handled by the `syntastic#c#GetLocList()` framework, but the special case of headers doesn't allow that.  How ugly.  Anyway, your patch discards the use of `g:syntastic_cuda_arch`, which would introduce a pointless backward incompatibility.  Please try commit f9eca22 instead.

nmoehrle:
Thank you! This is what I wanted and I think I learned a lot from your changes :-)

However, do you see a possibility to remove the dummy file (.syntastic_dummy.cu) afterwards? It doesn't feel right to clutter the directory.

Oh and correct me if I am wrong but the way it is set up now there is either the possibility for a configuration of arguments within the config file or within a vim config.
Since ```'args': syntastic#c#ReadConfig(g:syntastic_cuda_config_file)``` will be overridden by ```let g:syntastic_cuda_nvcc_args = ...```.
I personally would like to specify some global arguments (like '--std=c++11') in my ".vimrc" and some local arguments (like include paths) in the config file.

lcd047:
> do you see a possibility to remove the dummy file (.syntastic_dummy.cu) afterwards?

Sorry, but no.  Running `rm` automatically is always problematic, I don't want to go there.  Run a cleanup job from `cron` if you must.

> the way it is set up now there is either the possibility for a configuration of arguments within the config file or within a vim config.

Yes, this is intentional.  Use `args` if you have a few options.  Switch to a config file if you need more.  Use `autocmd`s to switch config files if you need different sets of options for different files.  That's the Vim way of <del>being a pain in the rear</del> doing things. :smile:

nmoehrle:
Would creating it in a temp directory be an option? Or is that to problematic due to different platforms?

Combining config file and arguments from the .vimrc works for gcc/clang, but I didn't figure out where in the code the combination actually happens. I just saw it in the debug output.

lcd047:
> Would creating it in a temp directory be an option?

You're trying to solve a non-problem.  Like I said, you have options to solve it on your side.

> Combining config file and arguments from the .vimrc works for gcc/clang, but I didn't figure out where in the code the combination actually happens.

`gcc` / `clang` use the `syntastic#c#GetLocList()` framework.

nmoehrle:
I am sorry to bother you again about this...

I still see the order in which arguments are overridden as a problem: If one starts out using ```args``` and later realizes that more options are required, thus creating a config file within the project folder, the config file will be ignored.
On the one hand this may lead to confusion and on the other hand I feel that local options (or in general options specified closer to the source) should not be overridden by global ones.

I don't know if it is currently possible to override the global (.vimrc) options if ```makeprgBuild``` is used. However, appending the options from the config file to ```args_before``` would resolve both these concerns and furthermore enable the workflow I was mentioning before (and already use with ```gcc``` / ```clang```).

lcd047:
> I still see the order in which arguments are overridden as a problem

Then feel free to fork `nvcc.vim`, rename the checker to something else, change it to suit your needs, and maintain it as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).  It's easy.

> I don't know if it is currently possible to override the global (.vimrc) options if `makeprgBuild` is used.

As noted in the manual, buffer-local variables `b:syntastic_cuda_nvcc_args` will override `g:syntastic_cuda_nvcc_args`.

> However, prepending the options from the config file to `args_after` would resolve both these concerns and furthermore enable the workflow I was mentioning before

So would adding extra options to `g:syntastic_cuda_nvcc_args_after` instead of `g:syntastic_cuda_nvcc_args`.

nmoehrle:
> As noted in the manual, buffer-local variables `b:syntastic_cuda_nvcc_args` will override `g:syntastic_cuda_nvcc_args`.

I meant from within the checker (overriding the `g:syntastic_cuda_nvcc_args` with the options from the config file).

> So would adding extra options to `g:syntastic_cuda_nvcc_args_after` instead of `g:syntastic_cuda_nvcc_args`.

I guess you mean `g:syntastic_cuda_nvcc_post_args` (since  `g:syntastic_cuda_nvcc_args_after` is internal to the checker) yes that could work.

PR #1786, opened 2016-06-06
Typescript tsc: use tsconfig.json if present

mrtracy:
This commit instructs the tsc compiler to search for a `tsconfig.json` file,
which indicates a typescript project. If found, tsc will be instructed to
compile the project instead of the individual file. If no `tsconfig.json` file
is present, single-file compilation will be used instead.

### DIFF
--- a/syntax_checkers/typescript/tsc.vim
+++ b/syntax_checkers/typescript/tsc.vim
@@ -39,9 +39,22 @@ function! SyntaxCheckers_typescript_tsc_IsAvailable() dict
 endfunction
 
 function! SyntaxCheckers_typescript_tsc_GetLocList() dict
-    let makeprg = self.makeprgBuild({
-        \ 'args': '--module commonjs',
-        \ 'args_after': (s:tsc_new ? '--noEmit' : '--out ' . syntastic#util#DevNull()) })
+    " Suppress compiler output; the necessary flag depends on the version.
+    let suppress_args = s:tsc_new ? '--noEmit' : '--out ' . syntastic#util#DevNull()
+
+    " Search upwards in the directory structure for a tsconfig.json, compiling
+    " the project (instead of the single file) if found.
+    let tsconfig = syntastic#util#findFileInParent('tsconfig.json', expand('%:p:h', 1))
+    if filereadable(tsconfig)
+        let makeprg = self.makeprgBuild({
+            \ 'fname': '',
+            \ 'args': '-p ' . fnamemodify(tsconfig, ':h'),
+            \ 'args_after': suppress_args })
+    else
+        let makeprg = self.makeprgBuild({
+            \ 'args': '--module commonjs',
+            \ 'args_after': suppress_args })
+    endif
 
     let errorformat =
         \ '%E%f %#(%l\,%c): error %m,' .


lcd047:
Please see #1778.  Sorry, but no.

mrtracy:
Any more info on why this dynamic check is bad? I see similar things being done in other compilers, such as the Java compiler (where I found the pattern in the first place). 

The current compiler does not provide a good experience for first-time vim users compiling a project; they will suddenly see compiler errors that are not present in any other common Typescript IDEs, which all support `tsconfig` files in this fashion.

lcd047:
1. Patches that change defaults are not accepted.
2. Patches that do things that can be easily achieved by changes to user's vimrc are not accepted, either.
3. The `javac` checker is a mess which is an order of magnitude more complex than any other checker.  It exists mainly for historic reasons, and it doesn't work particularly well with realistic-sized projects.  It should be simplified (or, better, killed), rather than duplicated elsewhere.
4. What `javac` does is a lot more involved than finding a config file.  Parsing `pom.xml` is not something that can be achieved by easy vimrc manipulations.
5. Fixing usability problems in external checkers is outside the scope of syntastic.
6. The main philosophy behind syntastic is that you, the user, are supposed to understand what you're doing, and syntastic shouldn't get in your way.  In most cases, making meaningful decisions automatically is not even possible, let alone desirable.
7. Keeping things backwards compatible is more important than making things smooth for first-time Vim users.  Frankly, first-time Vim users shouldn't use syntastic, to begin with.

I could go on, but you probably already disagree violently, so I'll stop here. :)

eggers:
@lcd047 So how is it handled if you have two projects with separate tsconfig's? Or, with an arbitrary number of `.d.ts` files specified? For example, I get these types of errors when using syntastic:

```
...|1 col 20 error| TS2307: Cannot find module 'jquery'.
...|2 col 20 error| TS2307: Cannot find module 'lodash'.
```

lcd047:
@eggers You creatively use `autocmd`s to set `args` to the appropriate combination of options for each project.  Something like this:
```vim
augroup some_name
    autocmd!
    autocmd BufRead,BufNewFile    */project1/*.ts
        \ let b:syntastic_typescript_tsc_args = '-p /path/to/config1.json' |
        \ let b:syntastic_typescript_tsc_fname = ''
    autocmd BufRead,BufNewFile    */project2/*.ts
        \ let b:syntastic_typescript_tsc_args = '-m es2015'
augroup END
```
Feel free to hate the Vim guys who came up with this way of doing things some ~20 years ago. :smile:

jnetterf:
@lcd047, I am grateful for all the thought that you have put into this. As a heavy TypeScript user and syntastic fan, I would like to offer my opinion.

When `tsc.vim` was originally written, [tsconfig.json](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html) did not exist, but it is now the de facto way of developing TypeScript projects. Searching for a parent tsconfig.json is the behavior implemented in Visual Studio Code, atom-typescript, tsuquyomi, and likely many other integrations. I understand that vim is not Visual Studio Code or Atom, and that vim users are expected to know what they are doing, however this is a good indication that the behavior provided by this patch is now expected by TypeScript users. By introducing and putting support behind tsconfig.json, Microsoft *changed* the expectation. In my mind, this makes a non-backwards compatible change worth considering.

I would also argue that searching for a parent tsconfig.json whenever a buffer is opened is not *trivial*. Many vim users (myself included!) are not anywhere close to proficient in Vimscript, but still find vim to be a huge productivity boost compared to other editors. You may see it as trivial, but I do not.

I'm writing this with the hope that you will reconsider. Either way, thanks again for all your work on syntastic!

@eggers and @mrtracy, in the meantime, using [tsuquyomi's syntastic integration](https://github.com/Quramy/tsuquyomi#integrate-with-syntastic) may work!

lcd047:
@jnetterf Did you happen to wonder what is the point of my list above, and how did I come up with it (and with the first two points in particular)?  At the time of this writing syntastic has 215 checkers for 102 filetypes.  In an ideal world, there would be 215 people that care about said checkers, or at least 102 people that care about said filetypes, and those people would maintain the corresponding checkers.  In the real world, you only get yours truly for the entire menagerie and the core.  Once in a while some kind soul ads a new checker, then runs away.  Seldom if ever anybody bothers to answer a question here about a checker they wrote, let alone post a patch to keep it up to date.  I'm afraid I value my sanity more than the expectations of the users of any of the individual 215 checkers.

Consequently, you have a few options:

* Spend your energy arguing with me about defaults until I remove the checker.  It happened several times in the past, and we lived happily ever after.
* Spend your energy convincing `tsc` maintainers to handle `tsconfig.json` automatically instead of relying on the editors out there to do it.
* Spend two minutes understanding my solution above with `autocmd` and friting a few lines in your vimrc.
* Maintain the `tsc` checker [externally](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).

Your choice.

jnetterf:
@lcd047 I understand where you're coming from -- syntastic is definitely an ambitious project, and it makes sense to limit the scope of accepted changes. Thanks again for maintaining it!

PR #1793, opened 2016-06-20
Add new C checker: ninja

rabinv:
Add a new C checker for the Ninja build system.  Currently, the only
difference between this and the make checker, apart from the name,  is
the fname, which uses Ninja's file^ syntax to build the first output for
the given source file.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -867,6 +867,22 @@ File containing project-specific options to be passed to "Splint"
 (such as defines or include directories), one option per line (cf.
 |syntastic-config-files|).
 
+------------------------------------------------------------------------------
+12. Ninja                                                  *syntastic-c-ninja*
+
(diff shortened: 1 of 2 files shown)


lcd047:
@rabinv 
> Currently, the only difference between this and the make checker, apart from the name, is the fname, which uses Ninja's file^ syntax to build the first output for the given source file.

Well, then changing the code is not needed, the same can be achieved by setting the options for the `make` checker:
```vim
let g:syntastic_c_make_args = ''
augroup syntastic_ninja
    autocmd!
    autocmd BufRead,BufNewFile * let b:syntastic_c_make_fname = syntastic#util#shexpand('<amatch>') . '^'
augroup END
```
Actually, the `make` checker itself is a kludge, since it can only detect errors in C rules.  It exists only for historic reasons, it isn't supposed to be copied elsewhere.  Sorry, but no.

PR #1795, opened 2016-06-21
prevent slow tern-lint execution

bounceme:
this linter runs best from the files directory

### DIFF
--- a/syntax_checkers/javascript/tern_lint.vim
+++ b/syntax_checkers/javascript/tern_lint.vim
@@ -30,6 +30,7 @@ function! SyntaxCheckers_javascript_tern_lint_GetLocList() dict
         \ 'makeprg': makeprg,
         \ 'errorformat': errorformat,
         \ 'preprocess': 'tern_lint',
+        \ 'cwd': expand('%:p:h', 1),
         \ 'returns': [0] })
 
     for e in loclist


lcd047:
What are you trying to do?  Whatever it is, using the `cwd` option is not going to work, `cwd` just changes the directory the checker is run from, it doesn't adjust the paths retrieved by `errorformat` accordingly.  Basically `cwd` is only useful for getting around broken checkers.

bounceme:
this checker is kind of broken, every js file gets parsed from the invoke location recursively.i'm talking to the creator to prevent this, but this is more accurate because tern can resolve dependencies defined in a js file

lcd047:
> this p.r is more accurate

Nope, this PR breaks the checker when you run `tern-lint` without a config file.  Not going anywhere from there.

The way I see it, there are two options forward:

1. You get things settled with `tern-lint`'s author, _then_ you post an update to syntastic.  This stands a chance to work only if the updated `tern-lint` accepts version queries, so that the checker can be enabled only when `term-lint` is new enough to make sense.
2. I remove the checker and you maintain it as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).

bounceme:
the config is made a default if no config file exists
https://github.com/angelozerr/tern-lint/blob/master/bin/lint#L25-L34
here is the globbing i'd like to prevent
https://github.com/angelozerr/tern-lint/blob/master/bin/lint#L147-L152

lcd047:
I'm sorry, your PR is broken regardless of `term-lint` internal details.

PR #1803, opened 2016-07-01
Support b:syntastic_filetype_map and b:syntastic_*_checkers

Rob--W:
This allows linters to be overridden in autocmd via buffer-local variables.

Buffer-local linting was already introduced in #532, but that only added support for flipping the linter setting for *all* types. This PR allows linting to be configured more granularly, per language and even with file type mappings. Here is an example of how these buffer variables can be used: https://wiki.mozilla.org/index.php?title=WebExtensions/Hacking&oldid=1138155#Vim

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -632,7 +632,12 @@ endfunction " }}}2
 
 function! s:_resolve_filetypes(filetypes) abort " {{{2
     let type = len(a:filetypes) ? a:filetypes[0] : &filetype
-    return split( get(g:syntastic_filetype_map, type, type), '\m\.' )
+    if exists('b:syntastic_filetype_map') && has_key(b:syntastic_filetype_map, type)
+        let type = get(b:syntastic_filetype_map, type)
+    else
+        let type = get(g:syntastic_filetype_map, type, type)
+    endif
+    return split(type, '\m\.')
 endfunction " }}}2
 
 function! s:_ignore_file(filename) abort " {{{2
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -199,6 +199,7 @@ function! g:SyntasticRegistry.getCheckers(ftalias, hints_list) abort " {{{2
     let names =
         \ !empty(a:hints_list) ? syntastic#util#unique(a:hints_list) :
         \ exists('b:syntastic_checkers') ? b:syntastic_checkers :
+        \ exists('b:syntastic_' . ft . '_checkers') ? b:syntastic_{ft}_checkers :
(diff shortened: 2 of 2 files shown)


lcd047:
I'm afraid you didn't put enough thought into this.  A given file has a well-defined `filetype`, and there is a single list of checkers associated to it.  That's why there a single `b:syntastic_checkers` for it.  Having 102 `b:syntastic_*_checkers` for 102 `filetypes`, out of which 101 are not used, doesn't really make sense.  Also, Vim's `autocmd`s provide enough flexibility to set `b:syntastic_checkers` by name, by language, by extension, and by many other criteria.  You just need to stop and think for a minute about it.

`b:syntastic_filetype_map` has a little more merit, it could be useful for being able to check a given file with a "foreign" checker (i.e. a checker for a different `filetype`).  This is not currently handled too well with syntastic, but you can still work around it if you must: you can temporarily change the `filetype` of the file before checking.  `autocmd`s to the rescue again.

Anyay, `b:syntastic_filetype_map` is not a good approach.  I'm actually working on a patch that would allow you to run _any_ checker as `<filetype>/<checker>`, regardless of the current `filetype` (I should have had it ready by now, but work happened).  I'll probably finish it some day.

Last but not least: your patch has other design problems, such as not using `syntastic#util#var()`.  Sorry, but no.

Rob--W:
Thanks for the quick feedback.

>  A given file has a well-defined `filetype`

This assumes that a file is given. I don't know the file, all I know is that a file path matches some pattern.

> you can still work around it if you must: you can temporarily change the filetype of the file before checking. autocmds to the rescue again.

Changing the ft affects many other things, including syntax highlighting.

>  I'm actually working on a patch that would allow one to run any checker as `<filetype>`/`<checker>`

This is indeed the problem that I had with `b:syntactic_checkers`, I was not able to use `eslint` with `autocmd FileType html`.

> Last but not least: your patch has other design problems, such as not using `syntastic#util#var()`

My current method effectively considers the union of the maps for mappings, if I were to use the function you proposed, then a local map clears anything. Either way works for me.

---

What do you suggest to accomplish my needs? I want to have specific linting rules for JS and HTML files within a specific subdirectory, and not affect the linting of any other file outside of this directory (so no global rules).

lcd047:
> Changing the ft affects many other things, including syntax highlighting.

Which is why I said you can change it _temporarily_.  Change the `filetype`, run checks, change the `filetype` back.

> I was not able to use `eslint` with `autocmd FileType html`.

Recent versions of syntastic have an `eslint` checker for `html`.

> What do you suggest to accomplish my needs?

Use `autocmd`s.  For path-based rules, there is little difference between global and local rules.  It's just that local rules are slightly more efficient.

Issue #1557, opened 2015-09-20
AlexJS Syntax Checker


chrisopedia:
**This is not an issue as it is more a question of help!!!**

I'm trying to write a text checker for [AlexJS](http://alexjs.com), which checks for insensitive writing.  Using the docs and the existing checkers I'm just not getting it to work.  I'm a noob when it comes to writing ViM, so any help would be awesome.  If this is not the appropriate place to post this, please direct me to it and I'll gladly move it over there.

```vim
"============================================================================
"File:        alex.vim
"Description: Syntax checking plugin for syntastic.vim
"Maintainer:  Newton Koumantzelis <chrisohpedia at gmail dot com>
"License:     This program is free software. It comes without any warranty,
"             to the extent permitted by applicable law. You can redistribute
"             it and/or modify it under the terms of the Do What The Fuck You
"             Want To Public License, Version 2, as published by Sam Hocevar.
"             See http://sam.zoy.org/wtfpl/COPYING for more details.
"
"============================================================================

if exists('g:loaded_syntastic_text_alex_checker')
    finish
endif
let g:loaded_syntastic_text_alex_checker = 1

let s:save_cpo = &cpo
set cpo&vim

" @TODO: figure out to highlight the word in question
" function! SyntaxCheckers_text_alex_GetHighlightRegex(item)
"     let term = matchstr(a:item['text'], '\m "\zs[^"]\+\ze"\($\| | suggestions:\)')
"     if term !=# ''
"         let col = get(a:item, 'col', 0)
"         let term = (col != 0 ? '\%' . col . 'c' : '') . '\V' . escape(term, '\')
"     endif
"     return term
" endfunction

function! SyntaxCheckers_text_alex_GetLocList() dict
    if !exists('s:alex_new')
        let s:alex_new = syntastic#util#versionIsAtLeast(self.getVersion(), [1])
    endif

    "<stdin>
    "   1:5-1:14  warning  `boogeyman` may be insensitive, use `boogey` instead
    "  1:42-1:48  warning  `master` / `slaves` may be insensitive, use `primary` / `replica` instead
    "  2:52-2:54  warning  `he` may be insensitive, use `they`, `it` instead
    "  2:61-2:68  warning  `cripple` may be insensitive, use `person with a limp` instead
    "
    "⚠ 4 warnings
    let errorformat =
        \ '%W<stdin>' .
        \ '%+C%p^%l:%c-%l:%c  warning  %m' .
        \ '%>%Z'

[...]

PR #1809, opened 2016-07-07
New checker for Markdown: alex

pixelastic:
This adds a text checker to markdown files using the [alexjs](http://alexjs.com/) linter. As they put on their website, Alex checks for `gender favouring, polarising, race related, religion inconsiderate, or other unequal phrasing.` Like warning of the use of `his`/`her` instead of `their`.

The linter expect the `alex` command to be available (`npm install -g alex`).

The tool seems quite young and might require a bit of configuration to avoid false positives, but this is out of the scope of this plugin. From my limited testing it seems to only output warnings, I did not manage to make it output errors (still, the code should handle just fine).

I used a `preprocess` for the parsing of the output, which is way easier than a pure `errorformat` but I'm pretty sure this can be handled by `errorformat` by anyone with better skills.

### DIFF
--- a/autoload/syntastic/preprocess.vim
+++ b/autoload/syntastic/preprocess.vim
@@ -8,6 +8,34 @@ set cpo&vim
 
 " Public functions {{{1
 
+function! syntastic#preprocess#alex(errors) abort " {{{2
+    let out = []
+    for err in a:errors
+      let split = split(err, '\s\+')
+
+      " Skipping header/footers
+      if len(split) < 4
+        continue
+      endif
+
+      let type = split[1]
+      if type ==# 'warning'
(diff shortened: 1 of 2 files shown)


lcd047:
This has been discussed in the past and rejected, please see #1557.

It's also no longer needed: syntastic has a checker for [textlint](https://github.com/textlint/textlint), which in turn has a plugin for `alexjs`.

Related, [proselint](https://github.com/amperser/proselint/) has a syntastic checker, and might add support for `alexjs` at some point.

pixelastic:
Oh sorry, didn't check the issues.

I never managed to make proselint correctly work, but I'll give textlint a try. Thanks.

PR #1810, opened 2016-07-08
Update MATLAB mlint after arguements

jasongraham:
Removes unexpected argument causing error from mlint.


lcd047:
Merged, thank you.

PR #1822, opened 2016-07-23
rubocop & foodcritic

jrieko:
rubocop
* refactor to postprocess function

foodcritic
* accommodate external rules not named `FC%n`
* `filterForeignErrors` so `g:syntastic_auto_jump` doesn't change the current buffer

### DIFF
--- a/syntax_checkers/chef/foodcritic.vim
+++ b/syntax_checkers/chef/foodcritic.vim
@@ -22,11 +22,12 @@ function! SyntaxCheckers_chef_foodcritic_GetLocList() dict
     let makeprg = self.makeprgBuild({})
 
     " FC023: Prefer conditional attributes: ./recipes/config.rb:49
-    let errorformat = 'FC%n: %m: %f:%l'
+    let errorformat = '%m: %f:%l'
 
     return SyntasticMake({
         \ 'makeprg': makeprg,
-        \ 'errorformat': errorformat })
+        \ 'errorformat': errorformat,
+        \ 'postprocess': ['filterForeignErrors']})
 endfunction
 
 call g:SyntasticRegistry.CreateAndRegisterChecker({
--- a/syntax_checkers/ruby/rubocop.vim
+++ b/syntax_checkers/ruby/rubocop.vim
@@ -25,26 +25,28 @@ function! SyntaxCheckers_ruby_rubocop_IsAvailable() dict
     return syntastic#util#versionIsAtLeast(self.getVersion(), [0, 12, 0])
 endfunction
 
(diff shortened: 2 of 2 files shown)


lcd047:
> rubocop
> * refactor to postprocess function

The function you're changing doesn't need refactoring, and `Postprocess` functions are intended for [external checkers](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external), they don't make sense for internal ones.  Please consider sending PRs only when they actually solve a problem, thank you.

> foodcritic
> * accommodate external rules not named `FC%n`

What problem does this solve?

> * `filterForeignErrors` so `g:syntastic_auto_jump` doesn't change the current buffer

You're supposed to use `syntastic_quiet_messages` to filter out errors you don't want to see, rather than disable said errors for everybody else.

Additionally, your PR changes three unrelated checkers at the same time.  Sorry, but no.

jrieko:
On Fri, Jul 22, 2016 at 11:41 PM, LCD 047 <notifications@github.com> wrote:

> rubocop
>
>    - refactor to postprocess function
>
> The function you're changing doesn't need refactoring, and Postprocess
> functions are intended for external checkers
> <https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external>,
> they don't make sense for internal ones. Please consider sending PRs only
> when they actually solve a problem, thank you.
>
Ok, gotcha. Withdrawn.

> foodcritic
>
>    - accommodate external rules not named FC%n
>
> What problem does this solve?
>
>
>    - filterForeignErrors so g:syntastic_auto_jump doesn't change the
>    current buffer
>
> You're supposed to use syntastic_quiet_messages to filter out errors you
> don't want to see, rather than disable said errors for everybody else.
>
How can `syntastic_quiet_messages` filter errors about files other than the
current buffer? Foodcritic's output is per cookbook so, otherwise, the loc
list contains errors from other files and `syntastic_auto_jump` can change
the buffer to another file in the cookbook.

> Additionally, your PR changes three unrelated checkers at the same time.
> Sorry, but no.
>
Huhwut? This PR only changes rubocop and foodcritic.

lcd047:
> How can `syntastic_quiet_messages` filter errors about files other than the current buffer?

One way to do it:
```vim
autocmd FileType * if(expand('<amatch>') =~# '\m\<chef\>')
    \|     let b:syntastic_chef_foodcritic_quiet_messages = { '!file:p': '\V' . escape(expand('%:p'), '\') }
    \| endif
```
> Huhwut? This PR only changes rubocop and foodcritic.

Oh.  Well, the idea was so bad, it felt like it extended its bad influence to more than two checkers. :)

PR #1837, opened 2016-08-08
Add new `fsc` checker which loads classpaths from files

emilyst:
This checker uses `fsc` the same way as the standard checker, but it allows `fsc` to pass the compiler's parse stage so that better linting is available. To do this, it needs to reference the dependencies needed to build a file. It can find `.classpath` files to reference for this purpose, which the checker searches for recursively from the directory of the file it's attempting to compile.

Don't know if this is up to snuff to be included officially, but I've been using this checker for most of a year with no real issue, so I thought I'd offer it.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -5442,6 +5442,7 @@ SYNTAX CHECKERS FOR SCALA                           *syntastic-checkers-scala*
 The following checkers are available for Scala (filetype "scala"):
 
     1. fsc......................|syntastic-scala-fsc|
+    1. fsc-improved.............|syntastic-scala-fsc-improved|
     2. scalac...................|syntastic-scala-scalac|
     3. Scalastyle...............|syntastic-scala-scalastyle|
 
@@ -5455,6 +5456,33 @@ Checker options~
 
 This checker is initialised using the "makeprgBuild()" function and thus it
 accepts the standard options described at |syntastic-config-makeprg|.
+
+------------------------------------------------------------------------------
+1. fsc-improved                                 *syntastic-scala-fsc-improved*
+
+Name:        fsc-improved
+Maintainer:  Emily St. <github@emily.st>
+
+This checker is an improvement upon the basic `fsc` checker in that it can
+successfully pass the parser stage of compilation, provided that the
+dependencies are added to a classpath file which the checker can find.
+
+Classpath files can be built by hand or automatically by tooling.
+
+Checker options~
+
+This checker is initialised using the "makeprgBuild()" function and thus it
+accepts the standard options described at |syntastic-config-makeprg|.
+
+Additionally:
+
+                              *'g:syntastic_scala_fsc_improved_classpath_file'*
+Type: string
+Default: ".classpath"
+Name of a file the checker should look for when trying to build the compiler's
+classpath. This allows you to add a project's dependencies, whether they be
+class files or directories with JARs within.
+
 
 ------------------------------------------------------------------------------
 2. scalac                                             *syntastic-scala-scalac*
--- a/syntax_checkers/scala/fsc-improved.vim
+++ b/syntax_checkers/scala/fsc-improved.vim
@@ -1,0 +1,90 @@
+"============================================================================
+"File:        fsc_improved
+"Description: Syntax checking plugin for syntastic.vim
+"Maintainer:  Emily St. <hello@emily.st>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
+"             it and/or modify it under the terms of the Do What The Fuck You
+"             Want To Public License, Version 2, as published by Sam Hocevar.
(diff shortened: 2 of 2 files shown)


lcd047:
You don't need to duplicate the existing checker just to change the defaults.  What you do can be achieved without code changes, like this (inspired by the [FAQ][faq]):
```vim
function! FindClasspath(where)
    let cpf = findfile('.classpath', escape(a:where, ' ') . ';')
    return cpf !=# '' ?
        \ [ '-classpath', join(readfile(cpf), syntastic#util#isRunningWindows() || has('win32unix') ? ';' : ':') ] :
        \ []
endfunction

let g:syntastic_scala_fsc_args = [
    \ '-Xfatal-warnings:false',
    \ '-Xfuture',
    \ '-Xlint',
    \ '-Ywarn-adapted-args',
    \ '-Ywarn-dead-code', 
    \ '-Ywarn-inaccessible',
    \ '-Ywarn-infer-any',
    \ '-Ywarn-nullary-override',
    \ '-Ywarn-nullary-unit',
    \ '-Ywarn-numeric-widen',
    \ '-Ywarn-unused-import',
    \ '-Ywarn-value-discard',
    \ '-deprecation',
    \ '-encoding UTF-8',
    \ '-feature',
    \ '-language:existentials',
    \ '-language:higherKinds', 
    \ '-language:implicitConversions',
    \ '-unchecked',
    \ '-d', ($TMPDIR !=# '' ? $TMPDIR : '/tmp') ]

augroup syntastic_fsc
    augroup!
    autocmd FileType scala let b:syntastic_scala_fsc_args =
        \ get(g:, 'syntastic_scala_fsc_args', []) +
        \ FindClasspath(expand('<amatch>:p:h', 1))
augroup END
```
Aside from that, there are a few problems with your patch:

* `syntastic#util#shexpand('%: p')` doesn't make sense, because of the space between `%:` and `p`
* `syntastic#util#tmpdir()` creates a temporary directory that is never removed
* `syntastic#util#findFileInParent(a:filename, syntastic#util#shexpand('%:p:h'))` does shell quoting at the wrong place.  Please see the snippet above for a better approach.

All this would have had more merit if it just added an option for `fsc` to load the classpath.  As it is, sorry, but no.

[faq]: https://github.com/scrooloose/syntastic#faqconfig

emilyst:
I thank you for the feedback. Being able to have what I want without an entire new checker is very appealing.

However, your suggestion doesn't work, for a number of reasons.

The biggest problem is that the `fsc` checker hardcodes the `args_after` (`'args_after': '-Ystop-after:parser'`) which prevents the compiler from passing the parser step of compilation. It is unclear to me if this can be overridden, and it is the entire reason I wrote an entirely new checker.

There also appear to be a few syntax issues with what you wrote, and it's unclear to me how to resolve all of them. For one, `augroup!` looks like it should be `autocmd!`. Second, the classpath file doesn't appear to get found and read in appropriately. Just to give an example, here's Syntastic's debug output with your suggestion:

```
syntastic: 302.313927: SyntasticMake: called with options: {'errorformat': '%E%f:%l: %trror: %m,%Z%p^,%-G%.%#', 'makeprg': 'fsc ''-Xfatal-warnings:false'' -Xfuture -Xlint -Ywarn-adapted-args -Ywarn-dead-code -Ywarn-inaccessible -Ywarn-infer-any -Ywarn-nullary-override -Ywarn-nullary-unit -Ywarn-numeric-widen -Ywarn-unused-import -Ywarn-value-discard -deprecation ''-encoding UTF-8'' -feature ''-language:existentials'' ''-language:higherKinds'' ''-language:implicitConversions'' -unchecked -d /var/folders/6h/yqzst33n5tq8l59bddxqjvkr0000gn/T/ -classpath '''' -Ystop-after:parser /Users/emilyst/Development/[omitted]'}
syntastic: 303.154529: system: command run in 0.840251s
syntastic: 303.154664: checker output: ['bad option: ''-encoding UTF-8''', 'fscbad option: ''-encoding UTF-8''', '  fsc -help  gives more information', 'error: source file ''/Users/emilyst/Development/everything/-encoding UTF-8'' could not be found', 'one error found', '']
```

Is there any way to override `args_after`?

lcd047:
> The biggest problem is that the `fsc` checker hardcodes the `args_after` (`'args_after': '-Ystop-after:parser'`) which prevents the compiler from passing the parser step of compilation.

Done: commit c73673e moves it to `args`.

> For one, `augroup!` looks like it should be `autocmd!`.

Right, fixed.

> checker output: ['bad option: ''-encoding UTF-8''',

Fixed.

> -classpath ''''

This looks like `readfile()` failing.  Whatever, I added more safety nets, please try again.

emilyst:
Thanks for https://github.com/scrooloose/syntastic/commit/c73673e0f33a6b00e927ff65f366ba25735ad70f. I'll give this all a shot later. Thanks for giving me your time.

emilyst:
For future readers, the suggested customizations do work (provided one changes `<amatch>` to `<afile>`, since the autocommand is a FileType one). Thanks again.

PR #1840, opened 2016-08-09
Add warning error format

emilyst:
Allows Syntastic to detect and display warnings.

I can also add a bit to allow overriding the `fsc` errorformat variable, too, if need be.


lcd047:
Merged, thank you.

PR #1845, opened 2016-08-17
Addtion of Perl 6 support

nxadm:
Hi,

This PR adds Perl 6 support to syntastic. In my tests,  the implementation seems to correctly catch errors and highlight the cultpit. The reaction from the Perl 6 community has been very positive.

If you accept the PR, I have a few question regarding future additions to the linter. While this syntax checker has been done very similarly as the one for Perl 5 (that I used as an example), this may  change in the future. I talked to Perl 6 core devs, and we may add error output in JSONto simplify the job of linters and other tools. I also would like to keep the Perl 6 support in line with the development of the language, it self very recent and the development very active. Do I just send new PR for future additions/bug fixes?

If there is something that need to be changed or if you have questions, feel free to contact me.
Regards,

Claudio


lcd047:
1. Did you read the [checker guide][guide], and in particular the section about [duplicate checkers][dupes]?
2. Can you please squash your 16 commits into a single one and re-send your PR?

[guide]: https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide
[dupes]: https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#duplicates

nxadm:
Hi lcd047,

I did see that section, but this is not a duplicate checker. I took the structure of the perl5 checker as a starting point, but they don't have much in common (the run an other binary and the errors are not related at all).

I will gladly squash the commits.

Regards,

Claudio

lcd047:
Please leave out the changes to the FAQ when you do.  Syntastic never had a checker for Perl 6, so there can't be any question about it no longer working. :)

nxadm:
Ok, I'll do that on the other PR (with a cleaner commit history as asked).

PR #1846, opened 2016-08-17
Perl 6 support

nxadm:
This PR puts the adaptions in one single commit. It is the cleanup of PR1845 (https://github.com/scrooloose/syntastic/pull/1845).

To clarify: Perl 6 is not the successor of Perl 5, but a sister language that does not share any code.

In the meanwhile I moved the changes to a new repo so it can be activated as an external plugin in the meanwhile (https://github.com/nxadm/syntastic-perl6).

I repeat: feel free to ask for any adaptations or clarification you esteem necessary.

Thank you,

Claudio

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -63,9 +63,9 @@ DocBook, Dockerfile, Dust, Elixir, Erlang, eRuby, Fortran, Gentoo metadata,
 GLSL, Go, Haml, Haskell, Haxe, Handlebars, HSS, HTML, Java, JavaScript, JSON,
 JSX, LESS, Lex, Limbo, LISP, LLVM intermediate language, Lua, Markdown,
 MATLAB, Mercury, NASM, Nix, Objective-C, Objective-C++, OCaml, Perl, Perl
-POD, PHP, gettext Portable Object, OS X and iOS property lists, Pug (formerly
-Jade), Puppet, Python, QML, R, Racket, RDF TriG, RDF Turtle, Relax NG,
-reStructuredText, RPM spec, Ruby, SASS/SCSS, Scala, Slim, SML, Solidity,
+6, Perl POD, PHP, gettext Portable Object, OS X and iOS property lists, Pug
+(formerly Jade), Puppet, Python, QML, R, Racket, RDF TriG, RDF Turtle, Relax
+NG, reStructuredText, RPM spec, Ruby, SASS/SCSS, Scala, Slim, SML, Solidity,
 Sphinx, SQL, Stylus, Tcl, TeX, Texinfo, Twig, TypeScript, Vala, Verilog,
 VHDL, VimL, xHtml, XML, XSLT, XQuery, YACC, YAML, YANG data models, z80, Zope
 page templates, and Zsh. See the [manual][checkers] for details about the
--- a/autoload/syntastic/preprocess.vim
+++ b/autoload/syntastic/preprocess.vim
@@ -89,11 +89,11 @@ function! syntastic#preprocess#dockerfile_lint(errors) abort " {{{2
                 call add(out, msg)
             endfor
         catch /\m^Vim\%((\a\+)\)\=:E716/
-            call syntastic#log#warn('checker dockerfile/dockerfile_lint: unrecognized error format')
+            call syntastic#log#warn('checker dockerfile/dockerfile_lint: unrecognized error format (crashed checker?)')
             let out = []
         endtry
     else
-        call syntastic#log#warn('checker dockerfile/dockerfile_lint: unrecognized error format')
+        call syntastic#log#warn('checker dockerfile/dockerfile_lint: unrecognized error format (crashed checker?)')
     endif
     return out
 endfunction " }}}2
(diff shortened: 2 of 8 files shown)


nxadm:
The new commits do the following:
- Rollback adaption in FAQ as asked.
- Add highlighting to for an specific kind of error (multi reason eject).

lcd047:
I merged your PR in a branch named [perl6](https://github.com/scrooloose/syntastic/tree/perl6), with this patch on top: 10d3df6.

The error messages produced by `perl6` are almost as hostile to parsing as those of `elixir`.  As a result, the checker is a colossal mess.  It might become usable some day, perhaps when `perl6` can output errors as JSON.  (On a side note: from the point of view of syntastic, a simple `/path/to/file.pl:line:column:message` would be preferable).  Until then, it needs a lot more testing.  So please feel free to test the `perl6` branch, let me know of any problems, and / or submit PRs.

nxadm:
Thank you, I agree.

I will test the changes (I think it lost the ability to find unfindable libraries),  send PR's as needed.
In the meantime, I will incorporate your changes into an external plugin and ask people to test it. At the same time, I'll check for the JSON support (the core devs are not against it, just time).

C.

lcd047:
> I think it lost the ability to find unfindable libraries

Hopefully it didn't.  You're relying on UTF-8 characters to figure out what's going on; that's unsafe for a number of reasons, and (if I read the code correctly) it doesn't work on Windows.  It also pollutes the final messages with useless junk.  I set the `RAKUDO_ERROR_COLOR` environment variable to an empty string to disable the ANSI codes, and this also occasionally changes the error format.  I tried to get all cases your parser covered, but I don't have a comprehensive set of files to produce all errors, and making them up from the Rakudo tests would involve spending more time on this than I can afford.

There is something that is lost from the old Perl 5 checker: the ability to detect warnings.  That was the point of running `perl -c` twice, the first time with `-X`, and the second time with `-Mwarnings`.  Your checker didn't do that properly, so I removed it.  Perhaps the future error output functions will be able to display both errors and warnings at the same time, without stopping early or producing duplicates. :smile:

Maintaining this as an external checker for a while and encouraging people to test it is a good idea.  As I said, collecting a set of samples that illustrate the various errors would be useful too.

For the purpose of future maintenance, it would be useful if Rakudo would return version strings that would allow me to tell that version X came before version Y, even across development versions.  The current output of `perl6 --version` is useful, but not particularly nice from that point of view.  Printing something like a commit `SHA-1` and nothing else (like Go does) would be disastrous.

One more thing: you can pass length information for highlighting purposes from the preprocessor to the main `GetLocList()` via the `%n` format (which is mostly useless otherwise).  You'd just have to set the `nr` field in the error list later to cancel it.  See f.i. the way the `dartanalyzer` checker does it.

Also, if I read the code correctly you could set `RAKUDO_ERROR_COLOR` to some special markers, then use those markers when parsing, to get better column and highlighting information out of it.  Hopefully the future error output functions will add proper column information instead of colours. :smile:

lcd047:
@nxadm Any update on this?  I'd like to make a release in a few days before the great trainwreck of async features, and I'd like to include this checker.

nxadm:
Hi,

Sorry for the dormant state. I talked to upstream about a more structured error output. In the meanwhile, Rakudo 2016.09 added JSON error output, so the syntastic error checking would be less fragile.

What I wonder is if --with non-beta Rakudo still being pretty recent and not that widespread-- if we need to keep the regex based syntax checker around or just concentrate on the latest releases. What do you think?

C.

lcd047:
> In the meanwhile, Rakudo 2016.09 added JSON error output

That's great news.

> What I wonder is if --with non-beta Rakudo still being pretty recent and not that widespread-- if we need to keep the regex based syntax checker around or just concentrate on the latest releases.

Syntastic doesn't have an "official" checker for `perl6` yet, and starting with one that is going to be useless in a few month is probably not a good idea.

The JSON output still seems pretty rough in many ways, and not completely stable.  For example this runs into an unhandled exception:
```sh
RAKUDO_EXCEPTIONS_HANDLER=JSON perl6 -c module_not_found.pm6
```
(`module_not_found.pm6` is the test file in your repo).  Thus, ironically, this means I won't include the checker in the forthcoming syntastic 3.8.

On the other hand, I'll move a few things around for the async feature.  So the `perl6` checker will have to wait some more. _shrug_

nxadm:
Deal.

I am in direct contact with the core dev that added the feature, so I can send targetted bug reports and/or patches.

Waiting seems like the best option for now. In the meanwhile, there is the plugin for people that need perl6 support now.

lcd047:
I committed a modified version of your patch to the [perl6 branch](https://github.com/vim-syntastic/syntastic/tree/perl6).  The result is actually worse than parsing plain text errors, mainly because of `perl6` itself.  In particular some messages are not completely translated to JSON, f.i. those produced by your test files `module_not_found.pm6`, `did_you_mean.pm6`, and `undeclared_routine.pm6` (note the embedded newlines in the `message` field).

Also, the column number is inferred from the `pos` field, since `column` is always `null`.  This almost works, i.e. Vim can infer a column position from it, but only if you don't have multibyte characters in your sources.  The problem is `pos` represents the offset of the error in terms of UTF-8 characters from the beginning of file, while Vim needs byte offsets (cf. `:h byte2line()`).  So basically column numbers are off as soon as you have accented characters in the lines preceding the errors.  In principle various work-arounds are possible, but they are too slow for any practical use.

Last but not least, the errors are output as a JSON object rather than a list.  This is odd, since it means there can't be more than one error of any given type.  Also it means syntastic must sort the errors by line after parsing.

Please use the [perl6 branch](https://github.com/vim-syntastic/syntastic/tree/perl6) for further work on this checker.

nxadm:
OK,

I'll look into it, fix what I can and if needed contact upstream.

C.

lcd047:
This is now merged in 026e09fb, thank you.

PR #1851, opened 2016-08-19
Fix highlight of '-----> eject errors' (removing utf8)

nxadm:
Hi lcd047,

I am going to your changes bit by bit while adding it to the external plugin and creating test cases.

This PR restores the highlight of '-----> eject errors' that points to the exact error in the code. I removed the utf8 match (made unnecessary by the unsetting RAKUDO_COLORS).

Regards,

C.

### DIFF
--- a/syntax_checkers/perl6/perl6.vim
+++ b/syntax_checkers/perl6/perl6.vim
@@ -46,14 +46,23 @@ function! SyntaxCheckers_perl6_perl6_IsAvailable() dict " {{{1
 endfunction " }}}1
 
 function! SyntaxCheckers_perl6_perl6_GetHighlightRegex(item) " {{{1
+    " Arrow-eject errors
+    let parts = matchlist(a:item['text'],
+        \'------>\s*\(.\{-}\)<HERE>')
+    if !empty(parts)
+        return '\V' . escape(parts[1], '\')
+    endif
+    " Default (catches also '^Can only use'
     let term = matchstr(a:item['text'], '\m''\zs.\{-}\ze''')
     if term !=# ''
         return '\V' . escape(term, '\')
     endif
+    "Undeclare routines and names
     let term = matchstr(a:item['text'], '\m^Undeclared .\+:\W\zs\S\+\ze')
     if term !=# ''
         return '\V' . escape(term, '\')
     endif
+    "Not found modules
     let term = matchstr(a:item['text'], '\mCould not find \zs.\{-}\ze at')
     return term !=# '' ? '\V' . escape(term, '\') : ''
 endfunction " }}}1


lcd047:
Please try to understand what my code does.  Adding the `----->` to the highlighting patterns is neither needed, nor useful (it's highlighting the prefix, not the actual error).

Take a look at how the `----->` is formed [here][rk].  The `$.pre` consists of the [last 40 UTF-8 characters of the prefix][pre].  You can't meaningfully infer the string to highlight from it, but you _can_ infer the column of the error, provided that the prefix is shorter than 40 UTF-8 characters.  That's what my code does -- except the first version didn't compute UTF-8 lengths correctly, and it didn't take into account the cutoff at 40 characters.  Commit da983de fixes that (and a few other issues).

[rk]: https://github.com/rakudo/rakudo/blob/4674f14b8c0181a458cfdc6fed0f131eba186f13/src/core/Exception.pm#L562
[pre]: https://github.com/rakudo/rakudo/blob/4674f14b8c0181a458cfdc6fed0f131eba186f13/src/Perl6/World.nqp#L4244-L4246

nxadm:
Hi,

Actually the regex in my PR does correctly infer the string to high lighten. It does so on my test files ( being added to the external plugin repo). 

The reason for this is that when you disable colours with RAKUDO_ERROR_COLOR, the utf-8 error symbols are replaced by regular ascii strings. In this case, the utf-8 eject symbol (⏏) I used as the end boundary of the regex is replaced by <HERE>:

```
$ perl6 -c arrow_error.pm6 ===SORRY!=== Error while compiling /home/claudio/Code/syntastic-perl6/t/arrow_error.pm6
Preceding context expects a term, but found infix = instead
at /home/claudio/Code/syntastic-perl6/t/arrow_error.pm6:2
------>     has $.var ==⏏ 1;
$ RAKUDO_ERROR_COLOR=0 perl6 -c arrow_error.pm6 
===SORRY!=== Error while compiling /home/claudio/Code/syntastic-perl6/t/arrow_error.pm6
Preceding context expects a term, but found infix = instead
at /home/claudio/Code/syntastic-perl6/t/arrow_error.pm6:2
------>     has $.var ==<HERE> 1;
```

This is the error code test:
```
class ArrowErrorTest {
    has $.var == 1;
}
```

C.

lcd047:
> Actually the regex in my PR does correctly infer the string to high lighten.

Sorry but it doesn't.  Here's what your code does with the test file:
2016-08-20-135634_238x119_scrot
It highlights the prefix, as expected.

Here's what my code does:
2016-08-20-135746_247x106_scrot
It highlights the position after the second `=`, which is exactly what the error message says.

It also detects the column of the error:
2016-08-20-135839_901x170_scrot
This is important, because you can jump to the exact position in the file fith `:lfirst` and friends.

lcd047:
More example that your code doesn't behave correctly: add a valid statement before the faulty one, say like this:
```perl6
class ArrowErrorTest {                                                                                          
    my $foo = 77; has $.var == 1;
}
```
Your code includes the previous (correct) statement in highlighting:
2016-08-20-142107_302x89_scrot

nxadm:
Thank you for your patience.

I merged your changes into the standalone plugin. I am ready to call for wider testing and ask people to add error cases.

I don't know if you would like to be added to the CREDITS. Please tell me if you do.

lcd047:
> I merged your changes into the standalone plugin.

A few comments on your plugin:

* [`g:syntastic_extra_filetypes`][ft] should be put in the initialisation section of your plugin, i.e. in a file `plugin/something.vim`.  If you put it in `syntax_checkers/perl6/perl6-latest.vim` it won't take effect until syntastic gets to look for `perl6` checkers for the first time, which kind of defeats the point of creating it in the first place.
* You probably shouldn't have the 
* Please don't revive [`g:syntastic_perl6_interpreter`][interp].  The equivalent `g:syntastic_perl_interpreter` exists for a single reason: because the first version of the `perl` checker used that name some 6+ years ago.  Standard variable `*_exec` can be used instead, and having non-standard variables all over the place makes it a pain every time I need to refactor the checkers.
* You have an [`echom`][echom] in the `IsAvailable()` function. :smile:
* Any particular reason for changing the [`IsAvailable()`][IsAvailable] function?  the variable `*_exec` (in your case `g:syntastic_perl6_perl6latest_exec`, since you named your checker `perl6latest`) works automatically: it looks up `perl6` in the `$PATH` by default, and it overrides it with something else if you point it to a full path.
* Normally you shouldn't need to remove [`Syntax OK`][ok], all patterns not matched explicitly should be removed automatically.  If the OK message shows up as an error than something else is broken.
* You don't need to define a [`Strwidth()`][strwidth], you can just call `s:_width()` directly, provided that you put its [definition][def] above the point where you call it (this is a Vim quirk).

> I don't know if you would like to be added to the CREDITS. Please tell me if you do.

I don't really have a preference either way.

[ft]: https://github.com/nxadm/syntastic-perl6/blob/21c32a948aed613ca8a2f7f8ed18eff5343d2a14/syntax_checkers/perl6/perl6-latest.vim#L43-L48
[interp]: https://github.com/nxadm/syntastic-perl6/blob/21c32a948aed613ca8a2f7f8ed18eff5343d2a14/syntax_checkers/perl6/perl6-latest.vim#L83-L86
[echom]: https://github.com/nxadm/syntastic-perl6/blob/21c32a948aed613ca8a2f7f8ed18eff5343d2a14/syntax_checkers/perl6/perl6-latest.vim
[IsAvailable]: https://github.com/nxadm/syntastic-perl6/blob/21c32a948aed613ca8a2f7f8ed18eff5343d2a14/syntax_checkers/perl6/perl6-latest.vim
[ok]: https://github.com/nxadm/syntastic-perl6/blob/21c32a948aed613ca8a2f7f8ed18eff5343d2a14/syntax_checkers/perl6/perl6-latest.vim#L110-L112
[strwidth]: https://github.com/nxadm/syntastic-perl6/blob/21c32a948aed613ca8a2f7f8ed18eff5343d2a14/syntax_checkers/perl6/perl6-latest.vim
[def]: https://github.com/nxadm/syntastic-perl6/blob/21c32a948aed613ca8a2f7f8ed18eff5343d2a14/syntax_checkers/perl6/perl6-latest.vim

lcd047:
I added highlighting for "Could not find module".  Commit b968bea.

nxadm:
Ok, thank you.

I am heading back home so I will add your fixes later (thx!). I discovered the 'Syntax OK' bug after I sent a call-for-testing, so I added quickly that block as a temporary workaround. I just removed the echom. A added a small note in the README.md with your nick.

I hope to get enough feedback to make the plugin stable, but I am counting on a more structured error output of rakudo. I opened a feature request and I am discussing it with core devs. In short, the discussion is of we need a quick solution (output errors as JSON) or a more generic system.

C.

lcd047:
I suspect JSON output wouldn't be substantially simpler than a generic system.  From a quick look at the code, the main problem would be going from adding bits of error information here and there, to collecting all information in a central place and formatting it.  Going from JSON to, say, XML would then be trivial: just write another function to format the same structure as XML.

Anyway, while you are at this: it would be useful to have column numbers for errors whenever applicable, and have these numbers as byte offsets from the beginning of line, rather than counts of UTF-8 characters.

lcd047:
One more encoding fix: commit 4c56378.

nxadm:
Thank you. I haven't had the time to merge the changes to the standalone plugin yet (very busy work schedule atm).

I have good new on the structured errors front: it's work in progress, e.g. https://github.com/rakudo/rakudo/commit/aaf7c3c

C.

PR #1859, opened 2016-08-31
update html-tidy5 instructions

tedder:
One link was dead, and it's been vastly modernized.


lcd047:
I just reorganised the FAQ (commit 631df93).  There seemed to be little need for a dedicated section about HTML5 these days, so I removed it completely.  All relevant information is already in the manual, and the original (HTML4) `html-tidy` is no longer relevant.  Thank you for the PR though.

PR #1863, opened 2016-09-03
Add column range matcher for Hdevtools

DanMeakin:
This pull request makes a minor change to the hdevtools syntax checker. New versions of hdevtools seem to use column ranges in some error messages. Rather than a message which starts `Filename.hs:155:11: error:`, new versions of hdevtools are outputting messages such as `Filename.hs:155:11-13: error:`.

These messages are not currently being picked-up by the existing syntax checker. This fixes this issue.


lcd047:
Then the same problem might affect the other rules.  Please try 1fec9e8 instead.

DanMeakin:
Nice, thanks! I wasn't confident that the same problem was happening in other circumstances so didn't want to make more changes than necessary. However, that looks like a much better solution.

PR #1875, opened 2016-09-21
Checker luac: Fix errorformat with customized exec.

Oberon00:
luac has the annoying habit of prepending its executable name to all
error messages (see http://www.lua.org/source/5.3/luac.c.html#fatal and
CTRL+F `progname=`). E.g. when using

     g:syntastic_lua_luac_exec = luac52

the `errorformat` would match nothing.

E.g. the file `main.lua` containing just `0` produces the output

    luac52: main.lua:1: unexpected symbol near '0'

when checked with `luac52 -p main.lua`.


lcd047:
Thank you. These guys must have seen the fun we have with parsing `pylint` and they got envious.  I committed a somewhat simpler fix: 90b19d1.

PR #1881, opened 2016-10-02
Analyze in clang_check

hank:
Adding support for enabling clang_check analyze, which produces completely different warnings in source code.  Makes this error show up in the editor properly:

```
test.c:13:14: warning: The left operand of '<' is a garbage value
    for(i; i < 10; i++)
```
capture
capture2

Perhaps the way the options are set up should be tweaked a bit, but nothing breaks and it works as expected if g:syntastic_clang_check_analyze is unset, set to 0 or set to 1.  Also, maybe it should be set up to use ReadConfig so c and cpp could have different options.

### DIFF
--- a/syntax_checkers/c/clang_check.vim
+++ b/syntax_checkers/c/clang_check.vim
@@ -22,11 +22,20 @@ if !exists('g:syntastic_c_clang_check_sort')
     let g:syntastic_c_clang_check_sort = 1
 endif
 
+if exists('g:syntastic_clang_check_analyze') &&
+   \ g:syntastic_clang_check_analyze == 1
+    let g:syntastic_clang_check_analyze_str = '-analyze'
+else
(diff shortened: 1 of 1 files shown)


hank:
Also, importantly, I have to set the following to get compilation databases to work with clang_check according to the docs:
```
let g:syntastic_c_clang_check_post_args = ""
```
Otherwise, I would have just used the config file or post_args to add the argument.

lcd047:
> Adding support for enabling clang_check analyze

How does this differ from the `clang_tidy` checker?

Also, this patches that just change the defaults are not accepted.

hank:
Ah I didn't realize clang_tidy was the intended way to use the analyzer.  The clang_check docs should say something to that effect, since both clang-check and clang-tidy support invoking the analyzer, but clang-tidy seems WAY more appropriate.  Thanks for the clarification!

PR #1883, opened 2016-10-03
Add ncverilog support for verilog and systemverilog

limelightful:
Modify from iverilog.vim to support ncverilog.
Add systemverilog syntax check.

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -88,6 +88,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'solidity':      ['solc'],
         \ 'sql':           ['sqlint'],
         \ 'stylus':        ['stylint'],
+        \ 'systemverilog': ['verilator'],
         \ 'tcl':           ['nagelfar'],
         \ 'tex':           ['lacheck', 'chktex'],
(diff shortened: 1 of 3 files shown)


lcd047:
This won't fly, mainly because there can be at most one checker that uses `syntastic#c#GetLocList()` for each filetype.  This is a design limitation.

limelightful:
If I change the ``syntastic#c#GetLocLis()`` name and all other function may conflict.
Would this be OK?

lcd047:
> If I change the `syntastic#c#GetLocLis()` name and all other function may conflict.

I'm afraid I can't parse this sentence.  What you _can_ do to address the problem is to rewrite the checker to use `makeprgBuild()` instead of `syntastic#c#GetLocLis()`.  It won't have as many features though.  Or you could adapt the existing `verilator` checker to support both `verilator` and `ncverilog`.

Alternatively, you could maintain this as an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).  Since you care about this checker, this would be the preferred approach.

limelightful:
Forgive my poor English.
I think I should use external checker instead.

PR #1884, opened 2016-10-04
Enabling Racket checker do work without running the code

dkvasnicka:
The change I propose uses `raco make` instead of just running the code. This might be a controversial change for some people because in case of smaller scripts running the code takes no time and can actually uncover more bugs. However that's not the case for every program and using raco is more universal -- will work for all files and does not carry the risk of blocking Vim (now that syntastic doesn't use Vim 8 async yet) by "checking" a program with long running time.
I also changed the error format to include all lines in Racket's error output. As it is the messages shown by Syntastic often arent informative enough and one must run the code again in DrRacket to get to the core of the problem.

### DIFF
--- a/syntax_checkers/racket/racket.vim
+++ b/syntax_checkers/racket/racket.vim
@@ -20,12 +20,12 @@ let s:save_cpo = &cpo
 set cpo&vim
 
 function! SyntaxCheckers_racket_racket_GetLocList() dict
-    let makeprg = self.makeprgBuild({})
+    let makeprg = self.makeprgBuild({'args' : 'make'})
 
     " example of error message
     "eval-apply.rkt:460:30: the-empty-environment: unbound identifier in module
     "  in: the-empty-environment
-    let errorformat = '%f:%l:%v: %m'
+    let errorformat = '%f:%l:%v: %m, %m'
 
     let loclist = SyntasticMake({
         \ 'makeprg': makeprg,
@@ -43,6 +43,7 @@ endfunction
 call g:SyntasticRegistry.CreateAndRegisterChecker({
     \ 'filetype': 'racket',
     \ 'name': 'racket',
+    \ 'exec': 'raco',    
     \ 'enable': 'enable_racket_racket_checker' })
 
 let &cpo = s:save_cpo


lcd047:
> The change I propose uses `raco make` instead of just running the code.

Well, that can be achieved without changes to the code, by setting `syntastic_racket_racket_exe` in your vimrc.  As a general rule, patches that only change the defaults are not accepted.

> I also changed the error format to include all lines in Racket's error output.

Sadly, error messages without line numbers are essentially useless to syntastic.

> As it is the messages shown by Syntastic often arent informative enough and one must run the code again in DrRacket to get to the core of the problem.

Yes: `racket` prints error messages in bits and pieces as they happen, rather than going through a centralised function.  The result is that said error messages are essentially unparseable.  Various solutions have been suggested in the past, but nobody stepped up to actually implement one.  As it is, the `racket` checker is one of the least useful ones in syntastic and a prime candidate for future deletion.  Including it in its current form has been a mistake. _shrug_

dkvasnicka:
> Well, that can be achieved without changes to the code
Well, the same could be said about the current defaults which are suitable for _less_ people compared to what I'm proposing. Shouldn't defaults be the most broadly usable settings?

> Sadly, error messages without line numbers are essentially useless to syntastic.
What I added doesn't remove line numbers. It still displays the vital line info (parsed and highlighted) and allows Vim to correctly display the error side marks on appropriate lines. It just displays the most important part of the messages, too. This is especially useful with Typing Racket which is completely at odds with this checker in its current form. If you use this checker, try it and see. Or I can upload screenshots showing the difference.

lcd047:
> Shouldn't defaults be the most broadly usable settings?

As I said, patches that change the defaults are generally not accepted.  This has nothing to do with usability.  It has to do with me being the only active maintainer left, and with me not having a clue about the vast majority of the 200+ checkers currently supported.  There are about as many opinions about what these defaults should be as ar users.  I have better ways to spend my time than updating defaults for each of the 200+ checkers every few days.  Also, this allows you to upgrade syntastic without worrying about breaking your setup.  This situation is of course less than ideal, but the resources I can afford to spend on syntastic are limitted. _shrug_

> What I added doesn't remove line numbers.

The very point of syntastic is to let you jump to the place of errors.  You can't do that if the error mesages don't have line numbers.  You also can't do that if you can't parse the relevant messages.  Currently syntastic can parse <20% of the error messages `racket` can produce.  Adding informational messages to the location list won't change that. _shrug_

PR #1888, opened 2016-10-06
Added golang build flag for massive speedup

apourchet:
Until now, writing to any golang file would run a go guild
command that would take up to 10 seconds during which the
editor would be essentially frozen. By using the -i flag,
any subsequent writes to that file will be almost instantaneous.


lcd047:
This is not needed.  It can be achieved without code chabges, by setting `g:syntastic_go_go_build_args`.  Patches that change defaults are not accepted; sorry, but no.

PR #1891, opened 2016-10-06
Add checker for elixir's credo.

squaresurf:
Why:

* Credo is a nice static code analysis tool for elixir.

This change addresses the need by:

* Add credo.vim to the elixir folder.

Side effects:

* Credo can be used with Syntastic when enabled.

### DIFF
--- a/syntax_checkers/elixir/credo.vim
+++ b/syntax_checkers/elixir/credo.vim
@@ -1,0 +1,65 @@
+" vim: set sw=4 sts=4 et fdm=marker:
+scriptencoding utf-8
+if exists('g:loaded_syntastic_elixir_credo_checker')
+    finish
+endif
+let g:loaded_syntastic_elixir_credo_checker = 1
+
+if !exists('g:syntastic_elixir_credo_all')
+  let g:syntastic_elixir_credo_all = 1
+endif
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_elixir_credo_IsAvailable() dict
+    call self.log('executable("elixir") = ' . executable('elixir'))
+    if !executable('elixir')
+        return 0
+    endif
+
+    call self.log('executable("mix") = ' . executable('mix'))
+    if !executable('mix')
+        return 0
+    endif
+
+    call self.log('syntastic#util#system("mix credo help") = ' . syntastic#util#system('mix credo help'))
+    if match(syntastic#util#system('mix credo help'), 'The task "credo" could not be found') == -1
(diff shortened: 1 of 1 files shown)


squaresurf:
This is my first PR for syntastic or vimscript so I am very open to suggestions on how to improve this PR.

lcd047:
Actually, `elixir` is one of the checkers I'm considering for removal after the next release.  It produces unparseable error messages, and generally it has oddities that make it unsuitable for use with syntastic.  It's also going to break when syntastic switches to async checking.  Credo has better error messages, but it still adds its own oddities to those of Elixir.  Among other things it insists to run in UTF-8 locales, and it produces useless binary characters in its output.  Vim doesn't cope well with that.  If you insist to make a syntastic checker out of it you should consider making it an [external checker](https://github.com/scrooloose/syntastic/wiki/Syntax-Checker-Guide#external).

Specific notes about your code:

* it doesn't have a license
* `g:syntastic_elixir_credo_all` is pointless, options can be added via `g:syntastic_elixir_credo_args`
* running `mix credo --version` is probably a better validity check than `mix credo help`
* Credo returns errors encoded as UTF-8 regardless of the current locale, so you need to convert them
* your `errorformat` should match the letters `[F]`, `[D]`, etc. rather than the arrows
* syntastic can't handle informational messages (`%I` etc.); this is a historic limitation
* you should probably mark Credo messages as "style".

squaresurf:
This is all great feedback! Thank you for your time @lcd047.

The limitations of the elixir checker is why I wrote the credo checker. To be very clear, are you saying that you do not want to merge such a PR even if I address your comments?

What do you mean by?
> you should probably mark Credo messages as "style".

lcd047:
> The limitations of the elixir checker is why I wrote the credo checker.

As far as I can tell they have pretty different scopes.  Can Credo deal gracefully with syntax errors?

> To be very clear, are you saying that you do not want to merge such a PR even if I address your comments?

I'll merge it if you get it to an useable form and if you insist.  Just beware that it's going to break pretty soon, and that I'll also remove it if I decide to remove the `elixir` checker. _shrug_

> What do you mean by?
>> you should probably mark Credo messages as "style".

Please see the [FAQ](https://github.com/scrooloose/syntastic#faqstyle).

squaresurf:
SInce the future of the elixir checker is unsure, I've decided to give neomake a try. Thank you so much for all your hard work @lcd047. Syntastic has been a faithful friend for many years, but neomake and neovim have async checkers as well as makers for credo.

PR #1900, opened 2016-10-18
Do not assume we are assembling to an ELF

RyuKojiro:
This disallows things like the org directive, which are perfectly normal in other formats. It's better to just not specify, and give the user the option to add their format to the *_args variable, if they want.


lcd047:
Merged, thank you.

PR #1908, opened 2016-11-06
Change install url in readme

artofhuman:


lcd047:
I think there are more occurrences, and not onlyn in `README`.

Unrelated: please see #1909.

lcd047:
Thank you.  Fixed in 7993f79.

PR #1928, opened 2016-12-19
Make passive mode truly passive when set via SyntasticToggleMode

parantapa:
Currently, if a filetype is marked active in g:syntastic_mode_map via active_filetypes,
switching to passive mode via SyntasticToggleMode will continue to make all buffers
whose filetypes were included in active_filetypes active.

The behavior I expected is that a switch to passive mode via SyntasticToggleMode
at runtime should override the starting configuration via g:syntastic_mode_map.

The current patch does just that.

### DIFF
--- a/plugin/syntastic/modemap.vim
+++ b/plugin/syntastic/modemap.vim
@@ -57,6 +57,7 @@ function! g:SyntasticModeMap.toggleMode() abort " {{{2
 
     if self._mode ==# 'active'
         let self._mode = 'passive'
+        let self._activeFiletypes = []
     else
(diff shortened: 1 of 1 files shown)


lcd047:
> Currently, if a filetype is marked active in g:syntastic_mode_map via active_filetypes, switching to passive mode via SyntasticToggleMode will continue to make all buffers whose filetypes were included in active_filetypes active.

This is the intended behaviour.  It's consistent with what happens if you start syntastic in passive mode.  It has been that way for a very long time, and you don't give any reason for changing it other than your expectation.  Sorry, but no.

Issue #1895, opened 2016-10-13
Working with multiple Vala files


arrufat:
Maybe this question is a bit silly, but how can I tell Syntastic that the classes and methods I'm using in one file are implemented in another one?
There are no includes in Vala, as far as I know, as all the files are passed to the compiler, which works everything out.

lcd047:
According to `:h syntastic-vala-valac`, set `g:syntastic_vala_modules` or add the list of modules to a `// modules:` comment?

arrufat:
@lcd047 I tried that (by writing the name of the file, with and without the extension) but it does not work...
In fact it breaks syntastic: no errors are reported any more.

lcd047:
> In fact it breaks syntastic: no errors are reported any more.

Please open your test file, set `g:syntastic_debug` to 3, run the checker, run `:mes`, and post the output.

arrufat:
So, I've added the file in the modules list as `syntext.vala` in the `main.vala` file.
As you can see in line 10, it is added to the valac command preceded by a `--pkg`.
I think there should be another special line at the top named `// files:` or `// sources:` containing the files that are passed directly to the Vala compiler. What do you think?
```
"main.vala" 238L, 7750C written
syntastic: 41.347268: g:syntastic_version = '3.7.0-237 (Vim 800, Linux, GUI)'
syntastic: 41.347640: &shell = 'bash' (!), &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellred
ir = '>%s 2>&1', &shelltemp = 1, &shellxquote = '', &autochdir = 0, &shellxescape = ''
syntastic: 41.348743: UpdateErrors (auto): default checkers
syntastic: 41.349682: CacheErrors: default checkers
syntastic: 41.351436: g:syntastic_aggregate_errors = 0
syntastic: 41.351924: getcwd() = '/mnt/data/photoocr/texts/Generate_Synth_Texts/vala'
syntastic: 41.353399: CacheErrors: Invoking checker: vala/valac
syntastic: 41.355777: SyntasticMake: called with options: {'errorformat': '%A%f:%l.%c-%\d%\+.%\d%\+: %t%[a-z]%\+: %m,%
C%m,%Z%m', 'makeprg': 'valac -C --pkg syntext.vala  main.vala'}
syntastic: 41.368631: system: command run in 0.012215s
[...]

PR #1932, opened 2016-12-21
Vala: add support for additional source files

arrufat:
This adds the possibility to specify needed sources to build the current
Vala file by adding `// sources: ` at the top of the file, in the same
way as `// modules: ` and `// vapi_dirs: `.
It was discussed in issue #1895, but I think it's handy. And optional.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -6499,6 +6499,12 @@ Default: unset
 Space-separated list of "source" files to be passed to the compiler, the
 extension ".vala" is automatically appended.
 
+                                                *'g:syntastic_vala_flags'*
+Type: string or array of strings
+Default: unset
+Space-separated list of flags to be passed to the compiler. The leading "--"
+are prepended automatically.
+
 Notes~
 
 If |'g:syntastic_vala_modules'| is unset, you can also specify a list of
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -6493,6 +6493,12 @@ Default: unset
 Space-separated list of "vapi" directories to be passed as "--vapidirs"
 arguments.
 
+                                                *'g:syntastic_vala_sources'*
+Type: string or array of strings
+Default: unset
+Space-separated list of "source" files to be passed to the compiler, the
+extension ".vala" is automatically appended.
+
 Notes~
 
 If |'g:syntastic_vala_modules'| is unset, you can also specify a list of
@@ -6501,7 +6507,11 @@ module to load for the current file by adding a special comment starting with
 
 If |'g:syntastic_vala_vapi_dirs'| is unset, you can also specify a list of
 "vapi" directories for the current file by adding a special comment starting
-with "// vapidirs:" and containing a space-delimited list of names.
+with "// vapidirs: " and containing a space-delimited list of names.
+
+If |'g:syntastic_vala_sources'| is unset, you can also specify a list of
+"source" files needed by the current file by adding a special comment starting
+with "// sources: " and containing a space-delimited list of names.
 
 ==============================================================================
 SYNTAX CHECKERS FOR VERILOG                       *syntastic-checkers-verilog*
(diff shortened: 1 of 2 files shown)


arrufat:
I've also added support for custom flags to get rid of annoying warnings when other libraries use
deprecated features, since we can add something like:
// flags: enable-deprecated

lcd047:
Well, the initial feature is already broken.  `:h search()` gives a pretty good idea why: the result depends on where the cursor is, matches are not accepted at cursor's position, search doesn't wrap around at end of file, multiple "magic" comments are not handled in any way, etc.  Less obviously, you need to deal with quoting explicitly.

On the other hand, non-standard mechanisms like this are always a pain to deal with when changing the checker code.  The fact that nobody has complained yet about the problems above seems to imply this feature is seldom used.  The same effect can be achieved with `autocmd`s, which, while clunky, _are_ the Vim way.  Also, your patch basically creates four copies of the same function, and that doesn't help either.  All of it _can_ be rescued with some refactoring, but I'm rather tempted to remove this feature altogether, rather than let it grow.

So I suppose what I'm asking here is, how badly do you want this?

arrufat:
I understand, my patch was a five minute hack to make it work.
I do not want to interfere with Syntastic development, although I find this feature rather useful. Otherwise I will always get errors when using Vim with Vala source files. I guess I can always fork and apply this patch on top of the master branch if you want to remove it. Do what you think it's better for the development of Syntastic :)

lcd047:
I committed a fix for the problems I described above: e1929878.  Sadly, adding the features in your patch the same way would be too convoluted.  Still, like I said, you can set `b:syntastic_vala_valac_fname` and `b:syntastic_vala_valac_args` from `autocmd`s.

arrufat:
Thanks for the refactoring. However, is this change supposed to work as before?
I am getting the following error (it seems to ignore the `// modules: gio-2.0` line):
```
syntastic: 39.541528: g:syntastic_version = '3.8.0-12 (Vim 800, Linux, GUI)'
syntastic: 39.541867: &shell = 'bash' (!), &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellredir = '>%s 2>&1', &shelltemp = 1, &shellxquote = '', &autochdir = 0, &shellxescape = ''
syntastic: 39.542665: UpdateErrors (auto): default checkers
syntastic: 39.543297: CacheErrors: default checkers
syntastic: 39.545071: g:syntastic_aggregate_errors = 0
syntastic: 39.545471: getcwd() = '/home/adria/Projects/syntext'
syntastic: 39.546412: CacheErrors: Invoking checker: vala/valac
syntastic: 39.549295: SyntasticMake: called with options: {'errorformat': '%A%f:%l.%c-%\d%\+.%\d%\+: %t%[a-z]%\+: %m,%C%m,%Z%m', 'makeprg': 'valac -C   src/syntext.vala'}
syntastic: 39.661764: system: command run in 0.111763s
```

lcd047:
> However, is this change supposed to work as before?

Mostly.  The magic comments are only searched in the first 100 lines of the current file.

> `&shell = 'bash' (!)`

On a side note: if this works at all, it's by accident rather than by design.  You're supposed to have a full path here, not just `bash`.

arrufat:
Even with the full path in `&shell` it still does not work. You can see the `called with options` line, where `valac` is called without any extra arguments, and the magic comment is at the top of the file. I might be missing something, but it used to work before.

lcd047:
Then it's a bug.  The full path in `&shell` is unrelated.  I'll have to look at it.

arrufat:
Thank you :)

lcd047:
Ok, please try commit 40ac6c46.

arrufat:
Thank you, it seems to work again now :)

arrufat:
Sorry for commenting again. When you said my patch was to convoluted, you meant even after your refactoring?
I made a fork and a patch here, it's only 2 lines now.
https://github.com/arrufat/syntastic/commit/20b961eaf4602b427e202a2cd8501e3034a9fa07

lcd047:
> When you said my patch was to convoluted, you meant even after your refactoring?

Yes.

> I made a fork and a patch here, it's only 2 lines now.

The problem with this is you're adding input files `args` variable.  Input files should only be added to `fname`.  Nothing breaks for now, but you're adding the implicit assumption that input files can appear anywhere, for all checkers.  That means it would no longer be able to set `fname` to an empty string when needed.  Basically you're adding technical debt for no reason. _shrug_

> The problem with autocmd is that each file needs different vala flags and sources, so, should I put a vim modeline for those files?

You can only set options from `modeline`s, not variables.  Yeah, `autocmd`s are ugly, but that's what Vim offers.  You could try to take advantage of defining variables ahead of time and using file patterns.

PR #1936, opened 2016-12-26
Fix bug when Syntastic hasn't been loaded as a script

Tav0:
- On-demand loading of plugin manager conflicted with getting Syntastic version.

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -228,6 +228,9 @@ function! SyntasticCheck(...) abort " {{{2
 endfunction " }}}2
 
 function! SyntasticInfo(...) abort " {{{2
+    if !exists('g:syntastic_version')
+        call s:GetSyntasticVersion()
+    endif
     call s:modemap.modeInfo(a:000)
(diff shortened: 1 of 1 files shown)


lcd047:
What problem does this solve?

lcd047:
Ping?

Tav0:
Hi @lcd047, sorry for the late reply. I use the [vim-plug option](https://github.com/junegunn/vim-plug#plug-options) `on` (on-demand loading) and I tend an error (pic below) every time I do `:SyntasticInfo`

screen shot 2017-01-04 at 5 07 15 am

Thus, this PR solves this problem.

The reason for this bug is in the code when loading Syntastic version, it is assuming Syntastic will be loaded when VIM starts up, which in my case isn't so I have made it so that it can still grab the constant variable of the version even if Syntastic wasn't loaded from the start.

lcd047:
Please try this commit instead: 9ed5763b.

Tav0:
Sure, thank you.

PR #1940, opened 2016-12-30
Elixir checker accept optional environment

tsubery:
The solution proposed [here](https://github.com/vim-syntastic/syntastic/issues/1826) wasn't working for. elixirc reported errors on many files that mix wasn't. It must have something to do with build environments.

With this patch we can set ```let g:syntastic_elixir_compile_env = { 'MIX_ENV': 'test' }``` and
 workaround [this issue](https://github.com/phoenixframework/phoenix/issues/1165) by building in a different environment.


lcd047:
The `elixir` checker is already broken beyond fixing, mainly because error messages produced by Elixir are way too hard to parse with `errorformat`.  So this checker is pretty high on the list of things to remove in a future release of syntastic.

In the mean time, you could achieve what you want without changes to the code, by using a script wrapper that sets whatever environment variables you want.  Thus, sorry, but no.

Issue #1937, opened 2016-12-26
syntastic: error: checker python/python returned abnormal status 1


danalsan:
Same symptoms as https://github.com/vim-syntastic/syntastic/issues/1131 but apparently different root cause

:SyntasticInfo output

```
Syntastic version: 3.8.0-10 (Vim 800, Linux)
Info for filetype: python
Global mode: active
Filetype python is active
The current file will be checked automatically
syntastic: 4.324299: system: command run in 0.230407s
syntastic: 4.331391: system: command run in 0.006002s
Available checkers: flake8 pep8 pycodestyle pyflakes pylint python
Currently enabled checkers: python flake8 pylint
syntastic: 4.332797: &shell = '/bin/bash', &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellredir = '>%s 2>&1', &shelltemp = 1, &shellxquote = '', &autochdir = 0, &she
llxescape = ''
```

When saving a py file with syntactic errors:

```
Flake8 found issues
syntastic: 61.174707: g:syntastic_version = '3.8.0-10 (Vim 800, Linux)'
syntastic: 61.174808: &shell = '/bin/bash', &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellredir = '>%s 2>&1', &shelltemp = 1, &shellxquote = '', &autochdir = 0, &sh
ellxescape = ''
syntastic: 61.175078: UpdateErrors (auto): default checkers
syntastic: 61.175305: CacheErrors: default checkers
syntastic: 61.175740: g:syntastic_aggregate_errors = 0
syntastic: 61.175876: getcwd() = '/home/dani/personal/botijo'
syntastic: 61.176323: CacheErrors: Invoking checker: python/python
syntastic: 61.176823: SyntasticMake: called with options: {'errorformat': '%E%f:%l:%c: %m', 'makeprg': 'python /home/dani/.vim/bundle/syntastic/syntax_checkers/python/compile.py ''''', 'env'
: {'TERM': 'dumb'}, 'returns': [0]}
syntastic: 61.195610: system: command run in 0.018481s
syntastic: 61.195761: checker output: ['Traceback (most recent call last):', '  File "/home/dani/.vim/bundle/syntastic/syntax_checkers/python/compile.py", line 11, in <module>', '    compile
[...]

Issue #1924, opened 2016-12-05
syntastic: error: checker python/python returned abnormal status 1


baif666:
When I use syntastic I meet this problem

Flake8 found issues
syntastic: error: checker output:
Traceback (most recent call last):
  File "/home/stone/.vim/bundle/syntastic/syntax_checkers/python/compile.py", li
ne 11, in <module>
    compile(open(argv[1]).read(), argv[1], 'exec', 0, 1)
IOError: [Errno 2] No such file or directory: ''
syntastic: error: checker python/python returned abnormal status 1

Could anyone help me solve this problem?

lcd047:
Please open your test file, set `g:syntastic_debug` to 33, run `:SyntasticCheck python`, run `:mes`, and post the output.  Please use cut & paste rather than screenshots.

baif666:
I run :mes after :SyntasticCheck python 

"my_Code/stuff/my_arg.py" 16L, 423C
syntastic: 1.631629: g:syntastic_version = '3.8.0-3 (Vim 704, Linux)'
syntastic: 1.631855: &shell = '/bin/bash', &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellredir =
 '>%s 2>&1', &shelltemp = 1, &shellxquote = '', &autochdir = 0, &shellxescape = ''
syntastic: 1.632579: UpdateErrors: python
syntastic: 1.632901: CacheErrors: python
syntastic: 1.633758: g:syntastic_aggregate_errors = 0
syntastic: 1.633936: $TERM = 'xterm'
syntastic: 1.634056: $PATH = '/home/stone/.pyenv/plugins/pyenv-virtualenv/shims:/home/stone/.pyenv/shims:/home/stone/.pyen
v/bin:/home/stone/perl5/bin:/usr/local/texlive/2015/bin/i386-linux:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin
:/bin:/usr/games:/usr/local/games:/usr/local/texlive/2015/bin/x86_64-linux'
syntastic: 1.634213: getcwd() = '/home/stone'
syntastic: 1.697550: system: command run in 0.051171s
syntastic: 1.697791: python/python: getVersion: 'python --version': ['Python 2.7.6', '']
syntastic: 1.698030: python/python: python version = [2, 7, 6]
syntastic: 1.698306: CacheErrors: Invoking checker: python/python
[...]

PR #1941, opened 2017-01-07
Fixed NerdTree being used as filename by the checker.

wsfreund:
Hi there,

first, thanks for the awesome package! Lately, I have being having issues when using syntastic together with NERDTree plugin. In some circumstances, when I saved the file, it was complaining that the file didn't exist, and this was due to the fact that it was using the NERDTree buffer name instead of the file.

I tried to track down what was happening today, and discovered that, for some reason, the BufWritePost was being called with @% set to the NerdTree, and because of this fname on the g:SyntasticChecker.makeprgBuild(opts) function was failing. Thus, I changed the fname value to be retrieved through syntastic#util#shexpand( expand('<afile>', 1) )), instead using '%'.

Hope it does not crashes other cases, but it seems to be working nice :D

### DIFF
--- a/plugin/syntastic/checker.vim
+++ b/plugin/syntastic/checker.vim
@@ -182,7 +182,7 @@ function! g:SyntasticChecker.makeprgBuild(opts) abort " {{{2
     let parts = []
     call extend(parts, self._getOpt(a:opts, basename, 'exe', self.getExecEscaped()))
     call extend(parts, self._getOpt(a:opts, basename, 'args', ''))
-    call extend(parts, self._getOpt(a:opts, basename, 'fname', syntastic#util#shexpand('%')))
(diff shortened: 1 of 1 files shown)


lcd047:
Sadly, your patch doesn't make sense when syntastic runs in passive mode.

Similar problems have been reported in the past (cf. #1131, #1924, #1937).  At this point, a bug in syntastic can't be ruled out completely but is highly unlikely.  Your approach to a fix is naive either way.  You don't provide enough information to track the actual problem, but I suggest you try to disable all other plugins aside from NERDTree and syntastic and see if you can still reproduce it.

wsfreund:
Ok, sorry, I had reasons to imagine that this patch was too naive. Either way, I won't have time until end of march to discover exactly what is producing the issue. As such, I think this can be closed till I can take a deeper look into this.

PR #1942, opened 2017-01-09
Checker c: look up config in the work directory

Dridi:
The GNU autotools have the concept of VPATH builds [1] with different
directories for the source tree (srcdir) and the build tree (builddir).
It also has the possibility to generate files and feed them variables
created at configure time that can be platform-dependant. Generating
Syntastic config files could for example include CFLAGS with desired
warnings enabled so that Syntastic could report them.

Example in a `configure.ac` file:

    AC_CONFIG_FILES([
        .syntastic_c_config
        Makefile
        ...
    ])

In the `.syntastic_c_config.in` template:

    @CPPFLAGS@
    @CFLAGS@

However, this doesn't work in VPATH builds because the lookup usually
starts inside $srcdir in the directory where the C file is located. It
will now fall back to using `getcwd()` if nothing was found and allow
Syntastic to find generated config files if vim was open from $builddir.

This change is made in the C checker instead of the utility function 
because other language stacks may expect different behavior when it 
comes to generated files. CMake has a similar out-of-source build [2] 
feature, it is even recommended. For the autotools, VPATH builds must
work during a `make distcheck`, which makes them more than recommended.

[1] https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html
[2] https://cmake.org/Wiki/CMake_FAQ#What_is_an_.22out-of-source.22_build.3F

### DIFF
--- a/autoload/syntastic/c.vim
+++ b/autoload/syntastic/c.vim
@@ -23,6 +23,9 @@ function! syntastic#c#ReadConfig(file) abort " {{{2
     " search upwards from the current file's directory
     let config = syntastic#util#findFileInParent(a:file, expand('%:p:h', 1))
     if config ==# ''
+        let config = syntastic#util#findFileInParent(a:file, expand(getcwd(), 1))
+    endif
+    if config ==# ''
         call syntastic#log#debug(g:_SYNTASTIC_DEBUG_CHECKERS, 'ReadConfig: file not found')
         return ''
     endif


Dridi:
So much for two lines of code, eh? As a side note I'd like to share a couple cents on what a better solution would be in my opinion, but that's a heavier change.

Instead of looking up the work directory as a fall-back, I would look both there and in the checked file's "directory and upwards". The reason for looking the workdir is explained in the previous commit, and for that reason I don't think it should look upwards too.

Instead of returning the first file, it would be great to return and process them all. This way `autoconf` or CMake could generate config files relevant to the current build environment, and in source directories one could override/specialize the config. For example, one could write software with lots of strict warnings for production code and disable some of them to allow a fault injector to break the type system somewhere in the test sources.

If your source tree is the same as the build tree, you'd expect the config file found a the root of the workdir to be be ignored/deduplicated when found again after the upwards lookup the checked file's directory. I also think the file list should be returned/processed backwards, to start from the most general to the most specific config.

This specialization logic could be applied to any type of checker.

I'm a big fan of Syntastic by the way, this is my first contribution and a good occasion to throw a big THANK for being so helpful on a daily basis! I'm running this patch locally, hoping to remove it with the next 3.8.0 release :)

lcd047:
> this doesn't work in VPATH builds because the lookup usually starts inside $srcdir in the directory where the C file is located.

It may not be obvious from the manual, but you can set `g:syntastic_c_config_file` to a full path, not just a filename.  You can also set it from an `autocmd`, if you need to do that for more than one project.

> Instead of looking up the work directory as a fall-back, I would look both there and in the checked file's "directory and upwards". 

According to `:h syntastic-config-location`:

> A configuration file is looked up in the directory of the file being checked, then upwards in parent directories.

Dridi:
> It may not be obvious from the manual,

I did read the manual, both for syntastic and vim functions I came across like `expand` and `findfile` (technically documented in `finddir`). If I rely on vim script to do the lookup, then for an up-to-date syntastic user who clones a project with generated config files it won't JustWork(tm).

>According to `:h syntastic-config-location`:
>
>>A configuration file is looked up in the directory of the file being checked, then upwards in parent directories.

Yes, I came across this one. It started because I was surprised that after moving from a hard-coded file to a generated one it stopped working (missing includes etc) and this is embarrassing that while I proposed this change I forgot to update the docs. I can push a docfix too if you are considering this patch.

lcd047:
> for an up-to-date syntastic user who clones a project with generated config files it won't JustWork(tm).

Syntastic has conflicting checkers for the same filetypes, it can't possibly "just work".

> I can push a docfix too if you are considering this patch.

If you need me to put it formally: what you suggest can be accomplished without code changes.  Also, your patch makes the results dependent on the current directory.  Sorry, but no.

Dridi:
> Syntastic has conflicting checkers for the same filetypes, it can't possibly "just work".

Good point, so that wouldn't be a regression anyway.

> Also, your patch makes the results dependent on the current directory.

Sorry for being persistent, but my vim usage without Syntastic already depends on the current directory. When I work in a VPATH, I need to open vim from the build tree if I want `:make` to be able to jump to errors in correct buffers. That's expected behavior.

lcd047:
> When I work in a VPATH, I need to open vim from the build tree if I want `:make` to be able to jump to errors in correct buffers. That's expected behavior.

Then you can also write an `autocmd` to set `g:syntastic_c_config_file` to a file in the current directory.  Problem solved.

PR #1944, opened 2017-01-13
erlang_check_file.erl: Add support for Erlang.mk

dumbbell:
Now, it first determines the compilation "profile" (are we building tests or the program itself?) and the build tool. Once it found it - either rebar when there is a `rebar.config` file, or Erlang.mk when there is a
`erlang.mk` file - it calls the appropriate function to query ERL_LIBS paths and compiler options.

The Erlang.mk function uses the `show-ERL_LIBS`, `show-ERLC_OPTS` and `show-TEST_ERLC_OPTS` Make targets to query those informations.

Signed-off-by: Gerhard Lazu <gerhard@lazu.uk>


lcd047:
Merged, thank you.

PR #1951, opened 2017-01-23
Redirect stderr to /dev/null in pylint --version output

### LINKED ISSUE #1950: pylint version parsing doesn't like the deprecation warning in pylint 1.6.5
pylint 1.6.5 seems to have started outputting a deprecation warning to stderr:

```
$ pylint --version
pylint 1.6.5,
astroid 1.4.9
Python 3.6.0 (default, Dec 24 2016, 08:02:28)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)]
This option 'required-attributes' will be removed in Pylint 2.0This option 'ignore-iface-methods' will be removed in Pylint 2.0
```

...that last line goes to stderr, but it seems to be read in by the pylint syntastic plugin and causes it to fail.

**`:messages`**
```
syntastic: error: checker output:
This option 'required-attributes' will be removed in Pylint 2.0This option 'ignore-iface-methods' will be removed in Pylint 2.0pylint 1.6.5,
astroid 1.4.9
Python 3.6.0 (default, Dec 24 2016, 08:02:28)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)]
syntastic: error: checker python/pylint: can't parse version string (abnormal termination?)
```

Xophmeister:
Avoids trying to parse deprecation warning issued by pylint
Fixes #1950

### DIFF
--- a/syntax_checkers/python/pylint.vim
+++ b/syntax_checkers/python/pylint.vim
@@ -32,7 +32,7 @@ function! SyntaxCheckers_python_pylint_IsAvailable() dict " {{{1
         " On new-ish Fedora it's "python3-pylint 1.2.0".
         " Have you guys considered switching to creative writing yet? ;)
 
-        let version_output = syntastic#util#system(self.getExecEscaped() . ' --version')
+        let version_output = syntastic#util#system(self.getExecEscaped() . ' --version 2>/dev/null')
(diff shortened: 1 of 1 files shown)


Xophmeister:
I note that this solution isn't very portable, but it works on POSIX-based platforms.

lcd047:
As I said in #1950, this is a problem in your `pylintrc`, not in syntastic.

Unrelated, your patch is wrong, because:
1. `2>/dev/null` doesn't work on Windows
2. killing `stderr` can hide real problems.

Sorry, but no.

PR #1953, opened 2017-01-27
[haxe] disable output during syntax check

jdonaldson:
This small change takes advantage of an argument that prevents haxe from generating output  (or executing external commands) during an otherwise normal compilation.

(edited : not new)


lcd047:
This seems to fix a long-standing problem. However the option `--no-output` seems to have been added in `haxe` 1.14.  There should be a a version check for that.

jdonaldson:
1.14 was released nearly 10 years ago.  I think that's not a good reason for a version check.

lcd047:
Ok, merged.  Thank you.

Issue #1649, opened 2015-12-26
Latest Flow error format not supported


0xR:
I am running the latest flowtype (0.20.1) and I get this error message on save:
`syntastic: warning: checker javascript/flow: unrecognized error format`

0xR:
Error messages look like this now:
```
hello-flow.js:3
  3:   return a.length;
                ^^^^^^ property `length`. Property not found in
  3:   return a.length;
              ^ Number
```

lcd047:
The error format didn't change, in fact syntastic runs just fine after that first failure.  What happens is more stupid than that: minor bit rot in `flow`, which exposes a design problem in Vim's API.

* syntastic runs `flow status --show-all-errors --json file.js`
* `flow` starts a new server and tells you about it, but the message to `stderr` doesn't end with a newline (this is the bit rot part)
* Vim lumps together `stdout` and `stderr` from `flaw`
* as a result, syntastic can't tell where the informational message ends, and where the JSON results begin:
```
$ flow status --show-all-errors --json file.js    
Launching Flow server for /home/lcd047/tmp        
Spawned flow server (child pid=2219)              
Logs will go to /tmp/flow/zShomezSlcd047zStmp.log 
Started a new flow server: -{"passed":false, ...}
```
To work around this, either start the server before running Vim:
```
$ cd /path/to/your/project && flow start
```
-- but you need to remember to do that every time.

Or set `*_exe` (__not__ `*_exec`) to `flow check`:
```vim
let g:syntastic_syntastic_javascript_flow_exe = 'flow check'
```
-- but that will slow down all checks.

Or tell `flow` developers to end the relevant message with a newline -- but they'll probably argue that `stderr` should be read by humans, not by Vim (and they would be right).

It's outside the scope of syntastic to work around flaws in Vim. _shrug_ I'll add a note to the wiki though.

JustinTulloss:
This is a bigger problem than just starting flow as if flow hasn't completed the check since your last save, it'll display an informational message and a progress spinner. It seems to cause the same problem.

lcd047:
@JustinTulloss Do you have proof that this is an actual problem, or are you just speculating?  `flow` shouldn't print a progress spinner if `stderr` is not pointing to a terminal, and it isn't when `flow` is run from Vim.

JustinTulloss:
Not just speculating, but it's the message, not the progress spinner that's the problem. From the syntastic debug log:
```
[...]

PR #1954, opened 2017-01-30
Improving flow check speed

rodrigopr:
Flow keep a server running when executed with `status` command which reduces time of future checks.


lcd047:
It's a little more complicated than that, see #1649.  Fortunately, `flow` has an option `--quiet` since version [0.34.0](https://github.com/facebook/flow/issues/2185).  Please see bd03a27c.

rodrigopr:
Thanks @lcd047 :)

PR #1958, opened 2017-02-06
[haxe] add warning indicator to errorformat

jdonaldson:
The haxe compiler can emit warnings in some cases.  This change configures the errorformat to handle that correctly.


lcd047:
Merged, thank you.

PR #1962, opened 2017-02-11
Riot .tag files eslint support

linuxenko:
There is a good syntax highlighting plugins for vim, such as `vim-riot`. But there is a problem with support of `.tag` files by linters and syntastic. This is a general support of riot file types for syntastic (based on javascript's eslint.vim file). 

The plugins for support `.tag` files for eslint i know:
 * `eslint-plugin-riot`
 * `eslint-plugin-peopleai` - mine, based on `eslint-plugin-riot`.

### DIFF
--- a/syntax_checkers/riot/eslint.vim
+++ b/syntax_checkers/riot/eslint.vim
@@ -1,0 +1,79 @@
+"============================================================================
+"File:        eslint.vim
+"Description: riot syntax checker - using eslint.
+"             based on javascript checker maintained by Maksim Ryzhikov.
+"Maintainer:  Svetlana Linuxenko <linuxenko at gmail dot com>
+"License:     This program is free software. It comes without any warranty,
+"             to the extent permitted by applicable law. You can redistribute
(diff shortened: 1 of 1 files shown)


lcd047:
This is no longer needed, with recent versions of syntastic you can run `eslint` from files of any type, like this:
```vim
:SyntasticCheck javascript/eslint
```
You can use `autocmd`s to emulate active mode and `check_on_open`.  Something like this:
```vim
autocmd BufReadPost *.tag SyntasticCheck javascript/eslint
autocmd BufWritePost *.tag SyntasticCheck javascript/eslint
```
But even if that was not the case, there is a [recommended way](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#duplicates) to deal with duplicate checkers.  And even if _that_ was not the case, `tag` files seem too obscure to include in syntastic.  It would qualify for an [external checker](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#external) though.

Sorry, but no.

PR #1973, opened 2017-03-02
Honour a global .rubocop.yml

mluts:
When a rubocop command receives a single file to check, it ignores
"Exclude" options in .rubocop.yml located in the project root directory.
This commit fixes that, so you don't see errors you don't expect to see.

"--force-exclusion" option is supported already for a long time
see https://github.com/bbatsov/rubocop/commit/eae27629dc6a5b27c5e90e8b32fd5914c6ae6cc2

### DIFF
--- a/syntax_checkers/ruby/rubocop.vim
+++ b/syntax_checkers/ruby/rubocop.vim
@@ -26,7 +26,7 @@ function! SyntaxCheckers_ruby_rubocop_IsAvailable() dict
 endfunction
 
 function! SyntaxCheckers_ruby_rubocop_GetLocList() dict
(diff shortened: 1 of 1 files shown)


lcd047:
This can already be achieved without code changes, by adding `--force-exclusion` to `g:syntastic_ruby_rubocop_args`.  Patches that change defaults are accepted only when there are _very_ strong reasons for them (as in fixing security problems).

In addition, your patch adds `--force-exclusion` to `args_after`, which makes it impossible to override by the user.  Sorry, but no.

mluts:
@lcd047 i would not open the pull request if i would know that it is solvable with `.vimrc`, thank you. Sorry, didn't read the docs, just have looked straight into the code

PR #1974, opened 2017-03-02
Add example showing how to set g:syntastic_ruby_mri_exec

subelsky:


lcd047:
I believe your addition is in direct contradiction to the two paragraphs immediately above it.  You're supposed to either (1) activate `rbenv` and run Vim from inside the environment.  Then you don't need to set `*_exec`.  Or (2) write a wrapper script that activates `rbenv`, and point `*_exec` to the _script_.  Then you don't need to care about `rbenv which ruby`.  Actually, pointing `*_exec` to the `ruby` executable instead of running from a virtual environment is explicitly not supported and doesn't work, that's what that section in the FAQ is supposed to address. _shrug_

Sorry, but no.

subelsky:
Thanks for the prompt (if unnecessarily dismissive/sarcastic) reply. So you're saying it's better to do:

```vim
silent let g:syntastic_ruby_mri_exec = substitute(system('~/bin/my_script'),'\n\+$','','')
```
where my script is something like:
```
rbenv which ruby
```
? easy change to make if that's the recommended way of doing things.

lcd047:
> Thanks for the prompt (if unnecessarily dismissive/sarcastic) reply. 

If you read my reply as dismissive / sarcastic, I'm sorry, that was not my intention.

The recommended way is to activate `rbenv`, run Vim, and leave `g:syntastic_ruby_mri_exec` unset.  Does this fail to work for you?

subelsky:
No worries, easy to misconstrue things over the Internet! The issue I was trying to solve is that I usually have a whole bunch of MacVim windows open, each one associated with a different Ruby project, some of which have different Ruby versions, so I want to use a different syntax check in each window. Would it make more sense for me to set `b:syntastic_ruby_mri_exec` to the value I listed, using an auto-command? I'm new to `vimscript` so I apologize if this sounds naive.

lcd047:
This won't work.  It isn't enough to run the executable version X to make sure modules for version X are loaded at run time instead of those for other versions, that's why things like `rbenv` and `rvm` exist in the first place.

So you really want to have an _environment_ for version X in a window, and an _environment_ for a different version Y in another window.  From the little I know about `rbenv` this is not possible, since `rbenv` just creates symlinks.  It's possible in `rvm` though, just activate the right version in each window.  Nothing of this have anything to do with syntastic though. _shrug_

PR #1975, opened 2017-03-04
Switch from JSCS to Eslint

azat-io:
JSCS was merged with ESLint. Look here: http://jscs.info/

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -373,7 +373,7 @@ project's root rather than the current directory...__
 
 A. You can set up an `autocmd` to search for the configuration file in the
 current directory and upwards, and add it to the checker's options when found.
-For example for `jscs`:
+For example for `eslint`:
 
 ```vim
 function! FindConfig(prefix, what, where)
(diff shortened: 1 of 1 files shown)


lcd047:
Right, but `eslint` already looks for config files in the current directory and upwards, and thus it doesn't have the problem that FAQ entry is supposed to address.  Also `.eslintrc` is supposedly [deprecated][deprecated] too.

Anyway, the point of that FAQ entry is to show how this kind of things can be done, rather that refer exclusively to `jscs` and its quirks.  I suppose the `jscs` checker can be removed from syntastic in a future release of syntastic (say in at least a year from now), so perhaps we should postpone finding a better example until then (provided that it still makes sense by then, that is).

[deprecated]: http://eslint.org/docs/user-guide/migrating-from-jscs#terminology

ferdnyc:
> Also `.eslintrc` is supposedly [deprecated](http://eslint.org/docs/user-guide/migrating-from-jscs#terminology) too.

Only in the sense that the typed `.eslintrc.json`, `.eslintrc.js`, `.eslintrc.yml` names are now preferred to the untyped `.eslintrc` name. Any JavaScript-format `.eslintrc` is still supported, it just needs to be named `.eslintrc.js`.

lcd047:
@ferdnyc Feel free to submit a patch that removes all references `jscs` and provides a better example in the README.

PR #1990, opened 2017-03-21
Add typescript/standard checker

othree:
Since ESLint supports TypeScript by customized parser.
Standard is also possible to lint TypeScript files.


lcd047:
This is no longer needed.  With recent versions of syntastic you can just add `javascript/standard` to `g:syntastic_typescript_checkers`, run `:SyntasticCheck javascript/standard` etc.

othree:
OK, that works.

PR #1993, opened 2017-03-25
Call uname without shell.

FriedSock:
I don’t see a need to have `uname` called within a shell, calling the
program directly using the list syntax speeds up the execution of his
line by a potentially huge amount - which is good as it is run during
startup.

With shell set to `/bin/bash --login` it can take a relatively long time
to load up a login shell. Below is the difference in time for this
line during startup before and after the change, this is nearly half
a second.

```
count  total (s)   self (s)

1   0.479970   0.000027         let g:_SYNTASTIC_UNAME = split(syntastic#util#system('uname'), "\n")[0]

1   0.002503   0.000025         let g:_SYNTASTIC_UNAME = split(syntastic#util#system(['uname']), "\n")[0]
```

may solve: https://github.com/vim-syntastic/syntastic/issues/91

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -56,7 +56,7 @@ if s:_running_windows
     let g:_SYNTASTIC_UNAME = 'Windows'
 elseif executable('uname')
     try
-        let g:_SYNTASTIC_UNAME = split(syntastic#util#system('uname'), "\n")[0]
+        let g:_SYNTASTIC_UNAME = split(syntastic#util#system(['uname']), "\n")[0]
     catch /\m^Vim\%((\a\+)\)\=:E484/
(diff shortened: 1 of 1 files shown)


lcd047:
Actually, the main point of running `uname` is to check that running programs in a shell works correctly, the output is only used for debugging, to save me an extra back and forth when asking people about their config.  Also, `system(['uname'])` is supported only since `systemlist()` was added, in Vim 7.4.248.

If your shell takes half a second to start I'd recommend using a (much) leaner program as Vim's `shell`, or at least as syntastic's `shell` (cf. `:h 'syntastic_shell'`).  The vast majority of people are not using any advanced shell features of the shell in that position, and this does, in fact, impact running linters every single time, not just Vim's startup.

Sorry, but no.

PR #2000, opened 2017-04-09
New checker: alf/sweet.

rveens:


doc/syntastic-checkers.txt
```diff
@@ -264,6 +265,53 @@ the usual 'g:syntastic_ada_gcc_<option>' variables. The only exception is
 executable.
 
 ==============================================================================
+SYNTAX CHECKERS FOR ALF 	                      *syntastic-checkers-alf*
+
+The following checkers are available for Ansible (filetype "ansible"):
```
lcd047: `s/ansible/alf/` I presume...

syntax_checkers/alf/sweet.vim
```diff
@@ -0,0 +1,70 @@
...
+
+if exists('g:loaded_syntastic_alf_sweet_checker')
+    finish
+endif
+let g:loaded_syntastic_alf_sweet_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_alf_sweet_IsAvailable() dict
```
lcd047: You don't need a function `IsAvailable()`.

syntax_checkers/alf/sweet.vim
```diff
@@ -0,0 +1,70 @@
...
+    if !exists('g:syntastic_alf_compiler')
+        let g:syntastic_alf_compiler = self.getExec()
```
lcd047: Please don't add pointless configuration variables.

syntax_checkers/alf/sweet.vim
```diff
@@ -0,0 +1,70 @@
...
+    endif
+    call self.log('g:syntastic_alf_compiler =', g:syntastic_alf_compiler)
+    return executable(expand(g:syntastic_alf_compiler, 1))
+endfunction
+
+function! SyntaxCheckers_alf_sweet_GetLocList() dict
+	" example: sweet -i=filename.alf 
+	" there must be no spaces between -i= and the filename string
+	let makeprg = self.makeprgBuild({
+				\ 'exe': self.getExec(),
```
lcd047: You don't need to set `exe`.

syntax_checkers/alf/sweet.vim
```diff
@@ -0,0 +1,70 @@
...
+                \ "fname": "-i=" . shellescape(expand("%", 1)) })
```
lcd047: I suggest using `syntastic#util#shexpand('%')` instead of `shellescape(expand("%", 1))`.

syntax_checkers/alf/sweet.vim
```diff
@@ -0,0 +1,70 @@
...
+    return SyntasticMake({
+        \ 'makeprg': makeprg,
+        \ 'errorformat': errorformat,
+        \ 'defaults': {'bufnr': bufnr('')} })
+endfunction
+
+call g:SyntasticRegistry.CreateAndRegisterChecker({
+            \ 'filetype': 'alf',
+            \ 'name': 'sweet',
+			\ 'exec': 'sweet'})
```
lcd047: You don't need an `exec`.

lcd047:
This looks highly specialised.  Please consider maintaining an [external checker](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#external) instead.

rveens:
Thanks for the helpfull comments. 
The language/tool is indeed highly specialised so i suppose you are right and maintaining an external checker is the way to go.
I will close the pull request.

rveens:
https://github.com/rveens/alf-vim repository now contains the external ALF checker plugin.

lcd047:
For what it's worth, this is still less than perfect.  First you need a load guard for `plugin/sweet.vim`, i.e. something like this near the top of the file:
```vim
if exists('g:loaded_sweet_plugin')
    finish
endif
let g:loaded_sweet_plugin = 1
```
Otherwise if for whatever reason the file is loaded more than once, you end up with `g:syntastic_extra_filetypes = ['alf', 'alf', 'alf', ...]`.

Second, `'exe': self.getExec()` in `makeprgBuild()` is actually harmful, because it doesn't do proper quoting of shell special characters.  It's highly unlikely to cause any harm (since nobody in his right mind has, say, `'` in the name of the directories in their `$PATH`), but it's still wrong.  The default value does the right thing, so just leave out that line.

PR #2003, opened 2017-04-17
Fix format option of dartanalyzer: `--machine` option is deprecated

yykamei:
`--machine` format is deprecated.
I changed to use `--format=machine` option instead of `--machine`.


lcd047:
Then a version check is also needed, perhaps like this: a882ecbc.

PR #2004, opened 2017-04-18
Fixing error parsing for tslint v5

KurtPreston:
Tested to work against tslint `3.15.0`, `4.1.0`, and `5.1.0`.

### DIFF
--- a/syntax_checkers/typescript/tslint.vim
+++ b/syntax_checkers/typescript/tslint.vim
@@ -26,12 +26,20 @@ function! SyntaxCheckers_typescript_tslint_GetLocList() dict
         let s:tslint_new = syntastic#util#versionIsAtLeast(self.getVersion(), [2, 4])
     endif
 
+    if !exists('s:tslint_major_version')
+        let s:tslint_major_version = self.getVersion()[0]
+    endif
+
     let makeprg = self.makeprgBuild({
-        \ 'args_after': '--format verbose',
         \ 'fname_before': (s:tslint_new ? '' : '-f') })
 
-    " (comment-format) ts/app.ts[12, 36]: comment must start with lowercase letter
-    let errorformat = '%f[%l\, %c]: %m'
+    " Example output:
+    " ts/app.ts[12, 36]: comment must start with lowercase letter
+    if s:tslint_major_version >= 5
+        let errorformat = 'ERROR: %f[%l\, %c]: %m'
+    else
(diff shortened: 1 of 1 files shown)


lcd047:
Please consider this fix for the parsing error: 0bfac455.

As for `--type-check`, changes to the defaults are only accepted if they fix catastrophic problems (such as security hazards).  You can achieve the same effect by setting `syntastic_typescript_tslint_args` from an `autocmd`.

KurtPreston:
@lcd047 Wow, quick turnaround, thanks!  `0bfac45` works great.

I took your suggestion and pulled the type check logic out into a separate `autocmd`-triggered function.  In case anyone finds this trying to enable type-checking, the gist is here:
https://gist.github.com/KurtPreston/e3b385a97ab0cea549c6b89eac8efa58

Only issue is I'm currently grabbing the tslint version via a `system` call, but I'd like to pull it from syntastic instead (in case it's using a different tslint).  Any help as to what command to run to get the tslint version number from syntastic would be appreciated.

lcd047:
> I took your suggestion and pulled the type check logic out into a separate `autocmd`-triggered function.

Err, what I had in mind is more along the lines of:
```vim
if !exists('b:syntastic_typescript_tslint_args')
    if !exists(s:type_check)
        let s:type_check =
            \ syntastic#util#versionIsAtLeast(syntastic#util#parseVersion(systemlist('tslint --version')[0]), [5]) ?
            \ '--type-check' : ''
    endif
    let b:syntastic_typescript_tslint_args = s:type_check
    if globpath(&runtimepath, 'autoload/tsuquyomi.vim', 1) !=# ''
        let b:syntastic_typescript_tslint_args .=
            \ ' -p ' . syntastic#util#shescape(tsuquyomi#tsClient#tsProjectInfo(@%, 0)['configFileName'])
    endif
endif
```
Put this in a file `~/.vim/after/ftplugin/typescript.vim` (or equivalent).

> Only issue is I'm currently grabbing the tslint version via a `system` call, but I'd like to pull it from syntastic instead (in case it's using a different tslint).

Well for one you'd use `syntastic#util#system()` instead of `system()`, in case syntastic uses a different shell. :smile: Beyond that, you'd have to reconstruct it from the `makeprgBuild()` variables.  Syntastic itself doesn't expose that kind of details.

Perhaps an introspection mechanism for checkers could be useful, but that's unlikely to happen any time soon.

PR #2009, opened 2017-04-22
PHPStan - new checker for php

przepompownia:
Currently it depends of phpstan/phpstan#07351e3 or later from dev-master.


lcd047:
Clean checker, thank you.

> Currently it depends of phpstan/phpstan#07351e3 or later from dev-master.

Then a version check is needed.  Sadly, a Git `SHA-1` is no good for that, since given two Git IDs there is no easy way to tell which one was issued before the other.  The `go` checker has the same problem.

Also, as a general rule syntastic only supports released versions of linters, mainly because keeping track of what's going on in development versions for (currently) 233 checkers would involve way too much work for the maintainer (i.e. for yours truly).

So perhaps ping me back when `--errorFormat raw` is supported in a main release and you have a proper version check for it?  Sorry for the inconvenience.

Unrelated: you might want to add a license to the checker.

przepompownia:
Thanks for conditional initial approval. 
https://github.com/phpstan/phpstan/issues/233#issuecomment-296440856
I wait for publishing a stable release. Currently, I encountered no problem related to the output from `phpstan`.

lcd047:
As I said, just let me know when it's ready.

Garbee:
It looks like the error checking level is hard-coded. Does this currently handle allowing the developer to use their own `phpstan.neon` config file? It isn't clear looking over the code changes proposed.

lcd047:
> It looks like the error checking level is hard-coded.

It isn't, you can override it by setting `g:syntastic_php_phpstan_post_args`.

That said, `--errorFormat raw` should be moved from `post_args` to `post_args_after`.

> Does this currently handle allowing the developer to use their own `phpstan.neon` config file?

Can't see why not.

syntax_checkers/php/phpstan.vim
```diff
@@ -0,0 +1,36 @@
...
+endif
+let g:loaded_syntastic_php_phpstan_checker = 1
+
+let s:save_cpo = &cpo
+set cpo&vim
+
+function! SyntaxCheckers_php_phpstan_GetLocList() dict
+    let makeprg = self.makeprgBuild({
+        \ 'args': 'analyse',
+        \ 'post_args': '--level=5 --errorFormat raw' })
```
lcd047: This should be:
```vim
        \ 'exe_after': 'analyse',
        \ 'args': '--level=5',
        \ 'args_after': '--errorFormat raw' })
```

przepompownia:
It works: I have tested it by modyfying parameters in local `.vimrc` with enabled `exrc` option. 

@lcd047 `args_before` and `args_after` seems to be undocumented. 

Thanks for patch and proposition of change.

lcd047:
@przepompownia They're not accessible to the user, so they aren't mentioned in the manual.  They are however documented in the [guide][guide].

[guide]: https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#makeprgbuildoptions

ondrejmirtes:
I just [released PHPStan 0.7](https://github.com/phpstan/phpstan/releases/tag/0.7), so you can depend on a stable version.

lcd047:
Merged, thank you.  Commit 5209bed1.

przepompownia:
Thanks to all contributors 👍

PR #2010, opened 2017-04-25
Ensure that php lint errors are reported to standard out

guywithnose:
If error_log has been set in php.ini this resets it to the default


lcd047:
Thank you.  Forced settings should go to `args_after` though.  Commit 8e82ea7b.

PR #2018, opened 2017-05-12
[gotype] Update argument for checking test file.

adamfaulkner:
This argument changed from `-a` to `-t` in the recent past. It needs to be updated in order to be able to check `*_test.go` files.

See documentation here: https://godoc.org/golang.org/x/tools/cmd/gotype

Making this update fixed the issue for me.

### DIFF
--- a/syntax_checkers/go/gotype.vim
+++ b/syntax_checkers/go/gotype.vim
@@ -21,7 +21,7 @@ set cpo&vim
 function! SyntaxCheckers_go_gotype_GetLocList() dict
     let buf = bufnr('')
     let makeprg = self.makeprgBuild({
-        \ 'args': (bufname(buf) =~# '\m_test\.go$' ? '-a' : ''),
+        \ 'args': (bufname(buf) =~# '\m_test\.go$' ? '-t' : ''),
         \ 'fname': '.' })
 
     let errorformat =


adamfaulkner:
CC: @lcd047 since it looks like you're the only person who's touched this file recently.

lcd047:
@adamfaulkner Please find out what "recent past" means and translate it to a version check.  Keep in mind that syntastic doesn't support development versions of checkers (in this particular case because there is no easy way to tell whether a given git ID came before a given release or the other way around).

adamfaulkner:
It looks like the change happened here:

https://github.com/golang/go/commit/359ca5ccc8b801d6ca8f8e417135436ccde00212

It was pretty hard to track down because the tool was moved from the tools repo to the standard library, modified, then copied back into the tools repo. 
Since most people will install gotype by running `go get golang.org/x/tools/cmd/gotype`, and that results in having the `-t` version, I believe the right approach is to use the `-t` argument here.

The version in go 1.8 also uses `-t`.

lcd047:
@adamfaulkner So we're running `gotype`, but we're supposed to check the version of `go` itself.  Hey, more reasons to drop support for `go` in favour of [vim-go][vim-go], how unexpectedly nice!

Anyway, I presume this means you're expecting me to do it for you.  Ok, commit fc7d4cea.

adamfaulkner:
Thanks for taking a look.

PR #2021, opened 2017-05-17
Extend hdevtools checker with a config_file option.

merijn:


lcd047:
The main point of config files as implemented by syntastic is to deal with `-I` includes.  However `hdevtools check` doesn't accept that syntax, so you made an almost duplicate of the standard config file reader.  Then you pretended in the manual your function does the same as the standard one.

Unrelated: there are also technical problems with your patch, such as config files being meant to work with the compiler framework, not with `mageprgBuild()`.

Still unrelated: patches that do things that can be accomplished with a few lines of Vim scripting are not accepted.  Sorry, but no.

merijn:
I'm sure there's a bunch of technical issues, as I don't really understand syntastic's design. This was simply the quickest/simplest way I could figure out to get something working.

Honestly, the config_file thing as here is probably overkill, but the main thing I find myself repeatedly missing with Syntastic is the ability to specify project/directory specific checker flags. Some kind of generic method for loading project/directory specific values for g:syntastic_<lang>_<checker>_args would solve that once and for all, but I honestly have no clue where something like that should go.

lcd047:
> Some kind of generic method for loading project/directory specific values for g:syntastic___args would solve that once and for all

The Vim way to do that is to use `autocmd`s to set the relevant variables.  In my opinion that isn't a great way of doing things, but it's what Vim has to offer.  In this particular case you could do something like [this item][faq] in the FAQ.  You could adapt it to read options from a file, or set `*_args` variables.

[faq]: https://github.com/vim-syntastic/syntastic#faqconfig

Issue #2036, opened 2017-07-04
syntastic#util#tmpdir uses s:fuzz but the function is actually s:_fuzz


Drako:
What the title says.

Sadly this breaks the kotlinc checker of [udalov/kotlin-vim](https://github.com/udalov/kotlin-vim).
Though it can be worked around by setting `g:syntastic_kotlin_kotlinc_delete_output = 0`.

lcd047:
Fixed by 75d2d98.

PR #2037, opened 2017-07-04
Fix for #2036: syntastic#util#tmpdir should work again

Drako:
I just replaced the occurrences of s:fuzz with s:_fuzz


lcd047:
Merged, thank you.

PR #2052, opened 2017-08-04
Make ruby FAQ section more idiomatic

lloeki:
The ruby section of FAQ has been updated, removing terms and references that were pythonic, as the equivalent of pyenv/venv is typically handled separately in ruby land.

Sadly *"by adjusting your vim configuration"* is a cop out. I initially wanted to include some vimscript example but I quickly hit a number of roadblocks. It would look something like this inspired by the JS one, but:

- The `--gemfile` arg does not exist for `bundle exec`... yet (it does for `install`)
- It should be applied to `bundle exec`, not `rubocop`, which AFAIK is not possible to do with the current syntastic variables (see additional note below).

```vim
let g:syntastic_ruby_rubocop_exe = 'bundle exec rubocop'

function! FindConfig(prefix, what, where)
    let cfg = findfile(a:what, escape(a:where, ' ') . ';')
    return cfg !=# '' ? ' ' . a:prefix . ' ' . shellescape(cfg) : ''
endfunction

autocmd FileType ruby let b:syntastic_ruby_rubocop_args =
    \ get(g:, 'syntastic_ruby_rubocop_args', '') .
    \ FindConfig('--gemfile', 'Gemfile', expand('<afile>:p:h', 1))
```

I wonder if there could/should be a completely generic notion of wrapper in syntastic that would help in wrapping things and passing args to the right place, so that the following could be conditionally and programmatically set by the user like `b:syntastic_*_*_args` on a custom case by case basis (e.g presence of `Gemfile` or not), something to the effect of:

```vim
" Gemfile and .rubocop.yml found via something like FindConfig
let b:syntastic_ruby_rubocop_wrapper = 'env BUNDLE_GEMFILE=/some/where/Gemfile bundle exec'
let b:b:syntastic_ruby_rubocop_args = '--config /some/where/.rubocop.yml'
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -436,16 +436,27 @@ scripts.
 
 __4.12. Q. How can I check scripts written for different versions of Ruby?__
 
-A. Install a Ruby version manager such as [rvm][rvm] or [rbenv][rbenv],
-activate the environment for the relevant version of Ruby, and install in it
-the checkers you want to use.  Set `g:syntastic_ruby_checkers` accordingly in
-your `vimrc`, and run [Vim][vim] from the virtual environment.
+A. Install a Ruby version manager such as [rvm][rvm],  [rbenv][rbenv],
+[chruby][chruby] or [ry][ry], activate the relevant version of Ruby, and
+install in it the checkers you want to use.  Set `g:syntastic_ruby_checkers`
+accordingly in your `vimrc`, and run [Vim][vim] from the environment or
+location where the relevant ruby version will be active.
 
-If you're starting Vim from a desktop manager rather than from a terminal you
-might need to write wrapper scripts around your checkers, to activate the
-virtual environment before running the actual checks.  Then you'll need to
-point the relevant `g:syntastic_ruby_<checker>_exec` variables to the wrapper
-scripts.
+If you're starting Vim from a desktop manager rather than from a terminal and
+depending on the version manager you use, you might need to write wrapper
+scripts around your checkers, to activate the relevant version of ruby before
+running the actual checks.  Then you'll need to point the relevant
+`g:syntastic_ruby_<checker>_exec` variables to the wrapper scripts.
+
+With rvm gemsets the above will be sufficient but if you rely on bundler and
+the Gemfile to lock the checker versions you may want to either make use of
+`RUBYGEMS_GEMDEPS='-'`, install the [rubygems-bundler][rubygems-bundler] gem,
+or instruct Syntastic to use `bundle exec`. In all cases you will need to make
+sure the linter is run from the correct directory or use the environment
+variable `BUNDLE_GEMFILE` for bundler to catch the proper Gemfile while also
+pointing the checker to its eventual configuration file (e.g `rubocop
+--config`). This can be done in a wrapper script or by adjusting your `vimrc`
+configuration.
 
 <a name="faqperl"></a>
 


README.markdown
```diff
@@ -436,16 +436,27 @@ scripts.
...
-the checkers you want to use.  Set `g:syntastic_ruby_checkers` accordingly in
-your `vimrc`, and run [Vim][vim] from the virtual environment.
-
-If you're starting Vim from a desktop manager rather than from a terminal you
-might need to write wrapper scripts around your checkers, to activate the
-virtual environment before running the actual checks.  Then you'll need to
-point the relevant `g:syntastic_ruby_<checker>_exec` variables to the wrapper
-scripts.
+A. Install a Ruby version manager such as [rvm][rvm],  [rbenv][rbenv],
+[chruby][chruby] or [ry][ry], activate the relevant version of Ruby, and
```
lcd047: Making environments that survive all the way to programs run from Vim is less than trivial, to say the least.  Did you actually check that `chruby` and `ry` solve the problem in the question on at least Windows, Mac OS X, and Linux?  If you didn't, I'd prefer to leave them out and stick with the proven known solutions.

Also, shouldn't `[chruby][chruby]` and `[ry][ry]` have corresponding URLs somewhere?

README.markdown
```diff
@@ -436,16 +436,27 @@ scripts.
...
+accordingly in your `vimrc`, and run [Vim][vim] from the environment or
+location where the relevant ruby version will be active.
+
+If you're starting Vim from a desktop manager rather than from a terminal and
+depending on the version manager you use, you might need to write wrapper
+scripts around your checkers, to activate the relevant version of ruby before
+running the actual checks.  Then you'll need to point the relevant
+`g:syntastic_ruby_<checker>_exec` variables to the wrapper scripts.
+
+With rvm gemsets the above will be sufficient but if you rely on bundler and
```
lcd047: Sorry, but this paragraph is confusing.  It has no motivation in the context of the question it's supposed to address. _shrug_

lcd047:
> I wonder if there could/should be a completely generic notion of wrapper in syntastic that would help in wrapping things and passing args to the right place, so that the following could be conditionally and programmatically set by the user like `b:syntastic_*_*_args` on a custom case by case basis (e.g presence of `Gemfile` or not)

As I said earlier: a _shell_ script.  Write it, and point `..._exec` (not `..._exe` as above) to it.  Trying to achieve the same in Vim is a lot more complicated, and almost guaranteed to break in a hard-to-debug way at some point.

lcd047:
Thank you.  Merged the first two paragraphs.  Commit a835b0ad.

lloeki:
👍 This is significantly less jarring, especially for newcomers.

README.markdown
lloeki: > Making environments that survive all the way to programs run from Vim is less than trivial, to say the least. Did you actually check that chruby and ry solve the problem in question on at least Windows, Mac OS X, and Linux?

You are right, this is a tricky problem, hence the plurality of tools. They do solve the problem: I personally use `chruby` and a coworker uses `ry` daily on the relevant platforms. Both are in wide enough use to be considered appropriate for the task, each one catering to different personal workflows or opinions:

- rvm overloads the shell's `cd` builtin
- rbenv generates and uses shimming shell wrappers for each gem executable script, doing away with the cd thing and leaving your env untouched, but the shims can get out of sync
- chruby leverages environment variables (mostly `PATH`, `GEM_*` and `RUBY_*`)
- ry leverages symlinks and doesn't touch anything else (env or otherwise), but thus acts globally

URLs for the record:

- https://github.com/postmodern/chruby
- https://github.com/jneen/ry

lcd047: I don't doubt `chruby` and `ry` do what they claim in a terminal.  What I'm asking is whether the effects of running them survive all the way to the linters run from Vim.  All methods you describe above incur limitations to syntastic and break some popular tricks related to changing `exe` or `exec`.

So, did you actually test them __with syntastic__, on at least Windows, Mac OS X (or some *BSD), and Linux?  Did you test them with different __shells__ (`ksh`, `bash`, `zsh`), and with different methods of __changing__ shells in Vim?

lloeki: > I don't doubt chruby and ry do what they claim in a terminal. What I'm asking is whether the effects of running them survive all the way to the linters run from Vim.

This was the usage I was referring to, we're both vim+syntastic+{ry,chruby} users.

> What I'm asking is whether the effects of running them survive all the way to the linters run from Vim.

Yes.

> All methods you describe above incur limitations to syntastic and break some popular tricks related to changing exe or exec.

I'm interested in learning how so. Does using `exe` or `exec` actively mangling exported `PATH`, `GEM_*` or `RUBY_*` vars? Does it change the behaviour of system symlinks? Otherwise env vars just flow through vim as is to ruby and the checkers, and symlinks are just symlinks.

> So, did you actually test them with syntastic, on at least Windows, Mac OS X (or some *BSD), and Linux? Did you test them with different shells (ksh, bash, zsh), and with different methods of changing shells in Vim?

He's using ArchLinux and Debian, and I'm using Mac OS X, ArchLinux and Debian. Neither of `rbenv` and `rvm` work on Windows (unless you use Cygwin or WSL) since they are both written in `bash`, but as soon as you make that enter the picture they all equally work. I use `bash` and `zsh`, my coworker uses `zsh` and `fish`. Changing shells within vim has no bearing on calling an external binary named `ruby` or `rubocop` available through `PATH`.

lloeki: Anyway, yhis is getting way overboard and I've got better things to do than arguing over this kind of trivialities, so, whatever.

lcd047: > Does using `exe` or `exec` actively mangling exported `PATH`, `GEM_*` or `RUBY_*` vars?

No, but several internals rely on the result of `executable()`.  This is not important in simple cases, but it can get really messy in more complicated setups (think `node`).  There has been an example of this causing problems a few weeks ago.

> Does it change the behaviour of system symlinks?

No, but symlinks take effect globally, not per project.

> Otherwise env vars just flow through vim as is to ruby and the checkers, and symlinks are just symlinks.

Nope.  Vim may mimic the behaviour of a shell in many respects, but it isn't a shell itself.  Perhaps surprisingly, `system()` in Vim is a really complicated mess.  This is well-known for Python's virtualenv, where various Vim plugins try to mess with the environment and run into a wall. _shrug_

PR #2069, opened 2017-08-31
Improve chktex format to handle the column number.

thepith:
Up to now `chktex` is used with the `-v1` switch so that the column number was not given in the output of `chktex`.

In this pull request the output format of `chktex` was changed to also give the column number and the `errorformat` is adapted accordingly.


lcd047:
Good call, thank you.  Merged, with this patch on top: 7dec5973.

Issue #2071, opened 2017-09-01
Support for .vue files


pixelastic:
Hello,

[Vue.js](https://vuejs.org/) default file extensions is `.vue` and usually looks something like that:

```vue
<template language="pug">
<!-- HTML/Pug content goes here -->
</template>

<script>
// Javascript content goes here
</script>

<style language="scss">
/* (S)CSS content goes here */
</style>
```

In addition to that, one can change the value of `language` to chose a specific flavor of HTML/JS/CSS.

I have syntastic set up to work with each filetype individually (`.pug`, `.js` and `.scss`) but fail to understand how to make it work with a file that contains several languages.

I tried to `set ft=pug.javascript.css`, but `pug-lint` tries to parse the whole file and throw errors from the JS part (`eslint` seems to behave better and only report issues from the JS part). Is there a way in syntastic to define delimiters to lint only parts of a file?

I found [pug-lint-vue](https://www.npmjs.com/package/pug-lint-vue) that correctly only output pug errors coming from the `template` part. The output is similar to `eslint` so should be easy to parse, but I don't understand how I should define it.

I'm willing to work on a PR, but I would need some pointers as the best way to organize it :)

lcd047:
> I have syntastic set up to work with each filetype individually (`.pug`, `.js` and `.scss`) but fail to understand how to make it work with a file that contains several languages.

You don't.  This is the famous template problem, and to this day it still doesn't have any reasonable solution (or at least I'm not aware of one).  Many other template checkers would benefit from solving this generically, but that seems remote at best.

In the mean time syntastic doesn't care about the contents of your files, it just runs linters.  So if you can find a linter that can only look at each section individually, you're done.  From your description `pug-lint-vue` seems to do that but only for `vue`.  For the other languages you have to either bear with useless error messages from the unspecialised linters, or check each section by saving it to a separate file (and perhaps use `make` to re-assemble sections back).

What _doesn't_ work is use something like `sed` to isolate sections for a given language and check the results with the unspecialised linters.  You do get back a list of errors, but you can't map them back to the original file. _shrug_

pixelastic:
Thanks for the detailed and quick answer,

So I have `pug-lint-vue` and `eslint` that are able to output only relevant errors from a `.vue` input. For CSS, seems like there's a PR for `stylelint` [on the way](https://github.com/stylelint/stylelint/pull/2596) so I'll deal with that later.

Moving forward, how should I proceed? 

[...]

PR #2072, opened 2017-09-01
Adding pug_lint_vue as first Vue.JS checker

pixelastic:
As per #2071, here is `pug-lint-vue` as the first checker for Vue.js (`.vue`) files. 

I did not add `eslint` as another default checker for `.vue` files. Not sure how to do it actually.

Took me a while to understand that I had to rename `pug-lint-vue` to `pug_lint_vue` when naming it inside of syntastic. Took me even longer to catch that `pug-lint-vue filename.vue` was not doing anything and that I had to force a `./filename.vue` (or an absolute path).

I've manually updated the documentation as well.


lcd047:
Thank you.  Merged, with this patch on top: 0858da63.

PR #2075, opened 2017-09-09
PHPStan - require version 0.8.5 or higher

przepompownia:
Version number is then correctly displayed by executables from composer.

https://github.com/phpstan/phpstan/issues/160#issuecomment-322040877


lcd047:
Merged, thank you.

PR #2079, opened 2017-09-18
Improvements for 'scalac' checker borrowing from 'fsc' checker

emilyst:
Since the `scalac` compiler and `fsc` compilers are identical (one simply wraps the other to keep it hot-cached in memory), improvements from one can be moved to the other. This pull request borrows a couple of changes from `fsc`'s checker and moves them into the `scalac` checker.

### DIFF
--- a/syntax_checkers/scala/scalac.vim
+++ b/syntax_checkers/scala/scalac.vim
@@ -25,6 +25,7 @@ function! SyntaxCheckers_scala_scalac_GetLocList() dict
 
     let errorformat =
         \ '%E%f:%l: %trror: %m,' .
+        \ '%W%f:%l: %tarning:%m,' .
         \ '%Z%p^,' .
         \ '%-G%.%#'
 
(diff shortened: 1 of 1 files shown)


lcd047:
Merged, thank you.

If `scalac` and `fsc` are identical perhaps one of the corresponding syntastic checkers can be dropped?

emilyst:
> Merged, thank you.

I appreciate the merge.

> If `scalac` and `fsc` are identical perhaps one of the corresponding syntastic checkers can be dropped?

I agree they are redundant. `fsc` is much faster than `scalac` after its initial check, however. I think the ideal case is the `scalac` checker were kept, and its command (as passed to `makePrg` were configurable. That's probably, in fact, already possible, but I regret I'm not familiar enough with Syntastic to know offhand whether it is or how practical it is.

lcd047:
You can set `g:syntastic_scala_scalac_exe` or `g:syntastic_scala_fsc_exe`, but it's probably easier to just switch between running `:SyntasticCheck scalac` and `:SyntasticCheck fsc`.

emilyst:
If that's configurable, then, as you say, then I regard the two checkers as truly redundant. I intend to use only the `scalac` checker from this point. I'll let you decide how to proceed, I suppose, with it.

lcd047:
Ok, then I'm inclined to keep both.  Removing one of them would introduce an incompatibility, without solving any real problem.

PR #2082, opened 2017-10-02
New checker for C#: csc

mynyml:
The current C# checker is based on the `mcs` compiler from the Mono project, which is being replaced by the [Roslyn C# compiler](http://www.mono-project.com/docs/about-mono/releases/5.0.0/#csc), whose executable is now called `csc`.

C# 7 code can only be parsed by the new compiler.

This checker is based on the existing `mcs` checker.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -985,6 +985,7 @@ SYNTAX CHECKERS FOR C#                                 *syntastic-checkers-cs*
 The following checkers are available for C# (filetype "cs"):
 
     1. mcs......................|syntastic-cs-mcs|
+    2. csc......................|syntastic-cs-csc|
 
 ------------------------------------------------------------------------------
 1. mcs                                                      *syntastic-cs-mcs*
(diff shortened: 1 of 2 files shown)


lcd047:
The `C#` checker in syntastic is a dead end, and it should probably be removed at some point.  The recommended alternative for the last few years has been to use [OmniSharp][OmniSharp] instead, which has a syntastic checker.  As I understand it, OmniSharp already supports Roslyn (among many other things).

[OmniSharp]: https://github.com/OmniSharp/omnisharp-vim

mynyml:
OmniSharp is python2, and my system is all python3. Setting it up turned out to be too painful to be worth it in my case, and basic Syntastic worked just fine for my needs. I think it provides a fair alternative.

lcd047:
Then you're in the ideal position for maintaining this as an [external checker](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#external).  Personally I have zero interest for C#, and I'd rather move to kill the existing internal C# checker, rather than add more of the same. _shrug_

Sorry, but no.

PR #2088, opened 2017-10-11
Fix systemverilog

steakknife:
vim at some point folded parts of `ft=systemverilog` into `ft=verilog`, while still setting the different filetype. syntastic was expecting `ft=verilog` for everything. Also, adds sane defaults verilator to use systemverilog instead of plain verilog from 1995. Just works OOTB.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -107,6 +107,7 @@ SYNTAX CHECKERS BY LANGUAGE                          *syntastic-checkers-lang*
     Solidity.................................|syntastic-checkers-solidity|
     SQL......................................|syntastic-checkers-sql|
     Stylus...................................|syntastic-checkers-stylus|
+    SystemVerilog............................|syntastic-checkers-systemverilog|
 
     Tcl......................................|syntastic-checkers-tcl|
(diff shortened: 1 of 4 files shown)


lcd047:
Please note that:

* Patches that add functionality that can also be achieved without code changes are not accepted;
* Changes to linter default options are accepted only if they address serious problems (f.i. security);
* Checkers [shouldn't be duplicated](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#duplicates).

With these in mind:

* Is there a need for _both_ `verylog` and `systemverylog` checkers in syntastic?  Is there still a use for `verylog` these days?  If there isn't, `verylog` checkers could be renamed to `systemverylog`.  Otherwise:
* Do `verylog` and `systemverylog` checkers have the same default options?  If they do: `:h 'syntastic_filetype_map'` and `:h :SyntasticCheck`.  Otherwise:
* Duplicate checkers [can use different options](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#duplicates).

lcd047:
Ping.

lcd047:
No feedback, no cry. :)

PR #2096, opened 2017-10-26
New checker for Solidity: solhint

sohkai:
Based off the existing solium and solc checkers.

The error format string is suboptimal (it captures everything as an error), but I'm not really sure how to fix it right now. `solhint` outputs messages in as `%f:%l:%c: %m [(Error | Warning)/<rule-name>]`, and I can't for the life of me figure out how to make an error format that can parse a message stuck in the middle of the line.

An example output:

```bash
contracts/Payroll.sol:29:3: Explicitly mark visibility of state [Warning/state-visibility]
```

--------

I'll ask solhint if it'd be possible to change it so the error state comes first, but just wanted to check if there was any magic to handle this that I wasn't aware of.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -6305,6 +6305,7 @@ The following checkers are available for Solidity (filetype "solidity"):
 
     1. solc.....................|syntastic-solidity-solc|
     2. Solium...................|syntastic-solidity-solium|
+    3. Solhint..................|syntastic-solidity-solhint|
 
 ------------------------------------------------------------------------------
 1. solc                                              *syntastic-solidity-solc*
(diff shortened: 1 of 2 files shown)


lcd047:
Please add an entry to the help.  Also please mark the checker as a "style" one.

> The error format string is suboptimal (it captures everything as an error), but I'm not really sure how to fix it right now. `solhint` outputs messages in as `%f:%l:%c: %m [(Error | Warning)/<rule-name>]`, and I can't for the life of me figure out how to make an error format that can parse a message stuck in the middle of the line.

You can't do it dirrectly, since `%m` has to be able detect the end of the string.  But there are a few workarounds:

* Use the `compact` formatter instead of `unix`.  This is not documented, but `solhint` seems to be using `eslint`'s formatters, and `compact` has been available for a while.
* If you don't care about the `<rule-name>` part, you can do something like this:
    ```vim
    let &errorformat =                                               
        \ '%f:%l:%c: %m [%trror/%\S%\+],' .                          
        \ '%f:%l:%c: %m [%tarning/%\S%\+]'
    ```
* Finally you can re-write the errors to something like this, which _can_ be parsed directly:
    ```
    contracts/Payroll.sol:29:3: W Explicitly mark visibility of state [state-visibility]
    ```
    That's what `preprocess` functions are for.

sohkai:
> Use the compact formatter instead of unix. This is not documented, but solhint seems to be using eslint's formatters, and compact has been available for a while.

Wow, I didn't notice this until you pointed it out. I've changed it so that the error format's the same as ESLint's. Thanks for pointing out `%\S` and `preprocess` out for me as well :).

lcd047:
Thank you.  Merged, with this patch on top: b0e48d0f.

PR #2103, opened 2017-11-04
Increase stylelint support for .less files

zanona:


lcd047:
Thank you.  Merged, with this patch on top: 81026249.

PR #2123, opened 2017-12-14
Use correct erlang include paths for rebar3 projects

russss:
Add a few more include paths for erlang projects built using rebar 3.
This allows syntastic to pick up dependencies and resolve include paths
in multiple-app rebar3 projects (assuming that the default `apps`
parent directory is used).

Existing erlang projects should be unaffected.

It's possible that this could be done slightly more elegantly by
shelling out to the new `rebar3 path` command, but this is quite slow in
our setup and may not be configured correctly in editing environments.


lcd047:
Merged, thank you.

PR #2126, opened 2017-12-21
Add yarac as yara rule syntax checker.

albb0920:
[YARA](https://github.com/VirusTotal/yara) is a widely used pattern matching tool in the malware research field.

This pull request adds support for checking YARA rule syntax by running the official `yarac` compiler.


lcd047:
Thank you.  Merged, with this patch on top: 88dd6567.

PR #2129, opened 2018-01-03
Enable column numbers and warning messages in mypy checker

pbasista:
The preprocessing of the mypy's output, as well as its errorformat,
have been changed, so that the column numbers
and message severity are properly included.

The 'note:' lines are not taken into account and are skipped.

### DIFF
--- a/autoload/syntastic/preprocess.vim
+++ b/autoload/syntastic/preprocess.vim
@@ -635,9 +635,16 @@ function! syntastic#preprocess#mypy(errors) abort " {{{2
     let out = []
     for e in a:errors
         " new format
-        let parts = matchlist(e, '\v^(.{-1,}):(\d+): error: (.+)')
-        if len(parts) > 3
-            call add(out, join(parts[1:3], ':'))
+        let parts = matchlist(e, '\v^(.{-1,}):(\d+):(\d+:)? (error|warning): (.+)')
+        if len(parts) > 5
+            " strip the last character (the colon) and increment by one
+            let parts[3] = string(str2nr(parts[3][:-2]) + 1)
+            let parts[4] = parts[4][0]
+            call add(out, join(parts[1:5], ':'))
+            continue
+        elseif len(parts) > 4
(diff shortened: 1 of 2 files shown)


lcd047:
This doesn't work with my version of `mypy`, so at least a version check is needed:
```
$ mypy where.py 
where.py:21: error: Cannot find module named 'geoip2'
where.py:21: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)

$ mypy --version
mypy 0.570-dev-b99f43fba341ea8d39c2521f7152ea422e023191
```

pbasista:
I have tried the exactly same version of `mypy` as you have mentioned, as well as the current master, as well as the currently latest version that is in pypi. All of them behave in the same way when it comes to reporting issue severity and column numbers.

The column numbers are disabled in your mypy configuration, so they will not be reported to syntastic either and column 1 will be indicated for every mypy error. Maybe it would be better to force enable the column numbers using a command line option `--show-column-numbers` for syntastic. Not sure.

What does not work with this patch is that the message severity does not get captured correctly by syntastic in case column numbers are not enabled. This is probably because of `errorformat`. I will look into that later this week. When column numbers are enabled, the syntastic captures all the fields correctly.

Did you notice any other issue?

lcd047:
`--show-column-numbers`  has been added in v0.4.5.  Please try 1b7ef51a (based on your patch).

lcd047:
On a side note: `len(matchlist(...))` is always 0 or 10, because of the way `matclist()` works.  The test `len(parts) > 5` is just a better-looking way of saying `len(parts) > 0`.  It still succeeds if say only 3 subexpressions match.

pbasista:
I have tried https://github.com/vim-syntastic/syntastic/commit/1b7ef51ab52f9f673e587691f3f028887d56302c and it seems to be working correctly. The column numbers are always present thanks to the added command line option and the warnings, when enabled, are included as well.

Thanks for implementing it.

Issue #2130, opened 2018-01-04
hlint unable to load .hs file inside Cygwin


chekkal:
Haskell: hlint unable to load .hs file inside Cygwin
hlint.exe doesn't understand the unix file path in Cygwin it need to be converted into windows path.
This prevent syntastic from displaying warning for Haskell files

lcd047:
Please explain what did you do, what did you expect to happen, and what happened instead.

chekkal:
Inside  Cygwin the plugin does display warnings for .hs files using hlint.exe

lcd047:
Please read `:h syntastic-shellslash`.  If that doesn't lead to a solution please actually explain what did you do, what did you expect to happen, and what happened instead.  Don't assume I can just look over your shoulder to see what you're doing.

chekkal:
Inside Cygwin enviromnent with the following config in .vimrc

> _map <Leader>s :SyntasticToggleMode<CR>
> 
> set statusline+=%#warningmsg#
> set statusline+=%{SyntasticStatuslineFlag()}
> set statusline+=%*
> 
> let g:syntastic_always_populate_loc_list = 1
> let g:syntastic_auto_loc_list = 0
> let g:syntastic_check_on_open = 0
> let g:syntastic_check_on_wq = 0_

Open a .hs file having some warnings. Execute :SyntasticCheck
Error: No errors are displayed for this file
Expected: The lines with error need to be highlited

chekkal:
> "file.hs" 1L, 6C
> syntastic: 10.926937: g:syntastic_version = '3.8.0-100 (Vim 800, CYGWIN_NT-6.3)'
> syntastic: 10.927183: &shell = '/bin/zsh', &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellredir = '>%s 2>&1', &shelltemp = 1, &shellxquote = '', &autochdir = 0, &she
> llxescape = ''
> syntastic: 10.927949: UpdateErrors: hlint
> syntastic: 10.928541: CacheErrors: hlint
> syntastic: 10.929565: g:syntastic_aggregate_errors = 0
> syntastic: 10.929973: getcwd() = '/cygdrive/s/training/cardano/cardano-sl'
> syntastic: 10.939242: CacheErrors: Invoking checker: haskell/hlint
[...]

PR #2131, opened 2018-01-04
hlint unable to load .hs file inside Cygwin #2130

chekkal:


lcd047:
No description of what this is supposed to fix and how = reject.

PR #2148, opened 2018-02-18
Fix go vet output parsing for go >= 1.10.

marcopaganini:
- Changed errorformat of go vet to include go build style messages in
  the parsing of go vet invocations.
- With go 1.10, govet apparently now runs a "go build" to collect
  information about the entire project. As a result, go vet will
  now output both go build and go vet errors. Previously, only
  go vet specific errors were reported.

### DIFF
--- a/syntax_checkers/go/govet.vim
+++ b/syntax_checkers/go/govet.vim
@@ -24,6 +24,7 @@ function! SyntaxCheckers_go_govet_GetLocList() dict
 
     let errorformat =
         \ '%Evet: %.%\+: %f:%l:%c: %m,' .
+        \ '%f:%l:%c: %m,' .
         \ '%W%f:%l: %m,' .
         \ '%-G%.%#'
 


lcd047:
So the difference between errors and warnings is now lost?

marcopaganini:
This is a good point. Apparently, yes as govet will report both :(

marcopaganini:
An alternative would be to use "go tool vet" (which behaves just like the old "go vet") but the Go documentation explicitly recommends not using it, unless we're doing development on "go vet" itself.

lcd047:
Ok, merged, thank you.  People should be using `vim-go` instead anyway.

marcopaganini:
Thanks! Hmmm, I use vim-go. You mean I shouldn't use SyntasticCheck, but
GoBuild (from vim-go) instead?

lcd047:
No, I mean `go` checkers are peculiar, and syntastic doesn't really care to bend over backwards to accomodate for these peculiarities.  Just don't use syntastic for `go`, it isn't really a supported language. _shrug_

PR #2149, opened 2018-02-19
Fix puppet 5 parser

jas01:
The new (not so new now) version of pupper (5) change a little the
parser output. Now it's something like

  Error: Could not parse for environment production: Syntax error at 'group' (file: /manifests/homedir.pp, line: 16, column: 5)

this bad fix allow syntastic to still working with puppet 5.

Currently puppet parser got a option (--render-as) to allow (well should) to create to ouput of the parser as json/yaml but still not working. Maybe some day...

### DIFF
--- a/syntax_checkers/puppet/puppet.vim
+++ b/syntax_checkers/puppet/puppet.vim
@@ -30,7 +30,8 @@ function! SyntaxCheckers_puppet_puppet_GetLocList() dict
         \ '%-Gerr: Try ''puppet help parser validate'' for usage,' .
         \ '%-GError: Try ''puppet help parser validate'' for usage,' .
         \ '%A%t%*[a-zA-Z]: %m at %f:%l:%c,' .
-        \ '%A%t%*[a-zA-Z]: %m at %f:%l'
+        \ '%A%t%*[a-zA-Z]: %m at %f:%l,'.
+        \ '%AError: Could not parse for environment production: %m (file: %f\, line: %l\, column: %c)'
 
     return SyntasticMake({
         \ 'makeprg': makeprg,


lcd047:
This looks like `puppet` is spilling its guts because of a parsing error, rather than producing a validation error.  Well, there are a few really bad checkers that should never have been added to syntastic to begin with, and `puppet` is one of them (cf. #1435).  Perhaps this is a good time to kill it, since `puppet-lint` can deal with syntax errors just fine.

jas01:
Le 19/02/2018 à 17:04:03+0000, LCD 047 a écrit
> This looks like puppet is spilling its guts because of a parsing error, rather
> than producing a validation error. Well, there are a few really bad checkers
> that should never have been added to syntastic to begin with, and puppet is one
> of them (cf. #1435). Perhaps this is a good time to kill it, since puppet-lint
> can deal with syntax errors just fine.

Yes I known but currently (maybe because I'm stupid) but I'm unable to make
puppet-lint check syntax error,

For example I just delete a "," at the end of a line and

with puppet parser
    [root@io manifests]$ puppet parser validate init.pp Error: Could not parse for environment production: Syntax error at 'Boolean' (file: /home/jas/Travaux/puppet/modules/apache/manifests/init.pp, line: 19, column: 3)

and puppet-lint

    [root@io manifests]$ puppet-lint init.pp
    [root@io manifests]$

but If I create a wrong indentation puppet parser don't say a word

    [root@io manifests]$ puppet parser validate init.pp

but puppet-lint will complain

    [root@io manifests]$ puppet-lint init.pp
    WARNING: indentatis not properly aligned (expected in column 21, but found it in column 23) on line 155
    [root@io manifests]$

Regards

--
JAS01

lcd047:
You miss the point.  If `puppet-lint` can't detect a syntax error that's a bug with `puppet-lint`.  If syntastic claims to support a linter that has been broken for 3+ years and made no progress in the mean time towards fixing the problem that's a bug with syntastic.

The goal here is not to provide a functional replacement for `puppet`.  The goal is keep syntastic maintainable. _shrug_

jas01:
Le 19/02/2018 à 17:59:29+0000, LCD 047 a écrit
> You miss the point. If puppet-lint can't detect a syntax error that's a bug
> with puppet-lint. If syntastic claims to support a linter that has been broken
> for 3+ years and made no progress in the mean time towards fixing the problem
> that's a bug with syntastic.
>
> The goal here is not to provide a functional replacement for puppet. The goal
> is keep syntastic maintainable. shrug

I understand that and respect that.

All I'm saying is puppet-lint is according to what they claim to be :

  Check that your Puppet manifest conform to the style guide

so....conform to *style*.

As I understand puppet-lint do not attempt to do the same thing as puppet
parser does.

Regards.

lcd047:
Ok, I suppose there are still e few people using this.  Thank you, merged, with this patch on top: b7b473b2.

PR #2153, opened 2018-02-26
[clang-tidy] [clang-check] Correctly pass args

LeonardKoenig:
Passing the arguments like it was done is not documented and seems to
even override any options, this means that clang is in fact only invoked
with said options, which not only suppresses warnings but also ignores
input from compile_commands.json etc. including include directories.

The correct way to pass arguments to clang is to use -extra-arg=<arg>.

### DIFF
--- a/syntax_checkers/c/clang_check.vim
+++ b/syntax_checkers/c/clang_check.vim
@@ -28,13 +28,12 @@ set cpo&vim
 function! SyntaxCheckers_c_clang_check_GetLocList() dict
     let makeprg = self.makeprgBuild({
         \ 'post_args':
-        \   '-- ' .
         \   syntastic#c#ReadConfig(g:syntastic_clang_check_config_file) . ' ' .
-        \   '-fshow-column ' .
-        \   '-fshow-source-location ' .
-        \   '-fno-caret-diagnostics ' .
-        \   '-fno-color-diagnostics ' .
-        \   '-fdiagnostics-format=clang' })
+        \   '-extra-arg=-fshow-column ' .
+        \   '-extra-arg=-fshow-source-location ' .
+        \   '-extra-arg=-fno-caret-diagnostics ' .
+        \   '-extra-arg=-fno-color-diagnostics ' .
+        \   '-extra-arg=-fdiagnostics-format=clang' })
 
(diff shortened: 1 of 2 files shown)


lcd047:

LeonardKoenig:
Eg. when having a `compile_commands.json` file it's completely ignored if one passes any flags via the old technique since that overrides any settings. To only make additional settings, one needs to use the `extra-arg` flag. This is exactly what we want as we want to have the same flags as for a usual build but also flags to disable color output etc.

LeonardKoenig:
Easy to reproduce example:

Have a c file include a header that's not in the standard include path, create a `compile_commands.json` file that specifies said include path. clang-check and clang-tidy would ignore said file, thus always complain about missing include dirs.

lcd047:
Have you considered reading the manual?  F.i. `:h syntastic-cpp-clang_check`.

lcd047:
* What is "correct", and how do you validate correctness?
* Does your patch work for all versions of `clang-check` where the existing code works?
* Can you point out a situation where the method described in the manual doesn't work?

LeonardKoenig:
This feature was added 5 years ago: https://reviews.llvm.org/D919?id=2275

The correct way is the way it does what it should. Currently, all compiler flags etc. are ignored because the `--` overrides. But we just want to *add supplemental* compiler flags and not replace all flags.

This way we even turn *off* some warnings that the user might want.

lcd047:
Can you point out a situation where the method described in the manual doesn't work?

LeonardKoenig:
It's not "not working" in the strictest sense of the word. It's more "not working as expected". It's also -- strictly speaking -- working as specified.

One can set the post_args to an empty string, **but this might mean that the checker output might change and become unparseable**; the DB will be read and applied.

lcd047:
> It's more "not working as expected".

Then perhaps what needs to be fixed is your expectation.  Defaults are there to be overridden, and you can do that without changing the code.

> One can set the post_args to an empty string, **but this might mean that the checker output might change and become unparseable**; the DB will be read and applied.

It's up to you to prevent that from happening, by passing the correct options to the linter. _shrug_

Issue #2164, opened 2018-04-02
[BUG] syntastic did not work on vim8 + macOS 10.13.4 for go code


arthurkiller:
syntastic did not highlight the go code and did not do check on :wq.

I haven't change any configuration.

lcd047:
> I haven't change any configuration.

Then you're describing the intended behaviour: syntastic is not supposed to work out of the box with Go files.   Please consider using [vim-go][vim-go] instead.  Or if you absolutely must use syntastic to check Go files read the manual and find out how to enable the Go checkers.  Feel free to re-open this if you still have problems after that.

[vim-go]: https://github.com/fatih/vim-go

arthurkiller:
> I haven't change any configuration.
This means it works properly yesterday, but I just update the macOS (yes, unfortunately), and my code syntastic just go off... @lcd047 

Sure, I'm using vim-go as I can. Could you plz give me some help or if I can give you any evidence on this problem (eg. How about open the debug mode on Syntastic?)

```
let g:syntastic_go_checkers=['go']
let g:syntastic_enable_signs=1
let g:syntastic_error_symbol='!!'
let g:syntastic_style_error_symbol='!!'
let g:syntastic_warning_symbol='??'
let g:syntastic_style_warning_symbol='??'
let g:syntastic_always_populate_loc_list=1
let g:syntastic_check_on_open = 1
let g:syntastic_check_on_wq=1
let g:syntastic_aggregate_errors=1
let g:syntastic_auto_loc_list=1 "disable the quickfix window
set statusline+=%#warningmsg#
set statusline+=%{SyntasticStatuslineFlag()}
set statusline+=%*
```

```
" vim-go settings
let g:go_get_update=0
let g:go_list_type="quickfix"
let g:go_highlight_functions=1
let g:go_highlight_methods=1
let g:go_highlight_fields=1
let g:go_highlight_types=1
let g:go_highlight_operators=1
let g:go_highlight_build_constraints=1
let g:go_fmt_command="goimports"
let g:go_fmt_options = {
            \ 'gofmt': '-s',
            \ }
let g:go_fmt_autosave=1
let g:go_fmt_fail_silently=1
let g:go_def_reuse_buffer=1
let g:go_def_mode='godef'
let g:go_template_autocreate=0
nmap <leader>b :GoBuild<CR>
nmap <leader>r :GoRun<CR>
nmap <leader>t :GoTest<CR>
nmap <leader>cs :GoCallstack<CR>
[...]

PR #2166, opened 2018-04-08
fix the go syntax checkers with default go build mode

### LINKED ISSUE #2165: Go checker --buildmode archive in go1.10
An improvement in 2015 to reduce check times using go1.7 relied upon a 'bug' in the go builder which has been 'fixed' in go1.10.1. (I believe it worked in 1.10, but can't confirm)

The --buildmode archive flag now throws an error on main packages: `no packages to build`

(It may not be a bug, but the new behavior matches the docs)

Commit which added the `--buildmode archive` flag as default: https://github.com/vim-syntastic/syntastic/commit/3c2e7e4ce37bc68b54f23c62efd0be291e6dc450

Discussed in issue https://github.com/vim-syntastic/syntastic/issues/1528

workaround:
`let g:syntastic_go_go_build_args=''`

arthurkiller:
fix #2165  #2164

### DIFF
--- a/syntax_checkers/go/go.vim
+++ b/syntax_checkers/go/go.vim
@@ -56,11 +56,11 @@ function! SyntaxCheckers_go_go_GetLocList() dict
     " compiled by `go build`, therefore `go test` must be called for those.
     if bufname(buf) =~# '\m_test\.go$'
         let cmd = 'test -c'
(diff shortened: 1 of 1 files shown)


arthurkiller:
also close this https://github.com/vim-syntastic/syntastic/issues/2165

arthurkiller:
this really necessary for those who use vim + syntastic + vim-go as editor.

PR #2185, opened 2018-06-08
Fix prospector checker (Python3 compat)

richey-v:
The Python 3.x version of prospector sends some junk to stderr.
Simply redirecting stderr to DevNull solves issue.

### DIFF
--- a/syntax_checkers/python/prospector.vim
+++ b/syntax_checkers/python/prospector.vim
@@ -31,7 +31,7 @@ endfunction
 
 function! SyntaxCheckers_python_prospector_GetLocList() dict
     let makeprg = self.makeprgBuild({
-        \ 'args_after': '--messages-only --absolute-paths --die-on-tool-error --zero-exit --output-format json' })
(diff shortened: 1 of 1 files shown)


lcd047:
This is a bad fix, because it can hide unrelated problems, leading to `prospector` failing silently.  Ideally this should be fixed on `prospector`'s side.  What exactly is the junk you want to skip?

richey-v:
This is the difference of output between Python version invocations of 
>   `python2 -m prospector example.py > /dev/null`

(nothing)
and
> `python3 -m prospector example.py > /dev/null`
>   `Using config file /home/richey/.pylintrc`

And prospector itself is not the issue... rather its invocation of pylint, illustrated thus:
>   `python2 -m pylint example.py > /dev/null`

(nothing)
> `python3 -m pylint example.py > /dev/null`
>   `Using config file /home/richey/.pylintrc`

I may or may not pursue fixing this upstream with pylint.

lcd047:
As noted in the [FAQ][FAQ], the easy way to deal with this is to run Vim from `virtualenv`.

[FAQ]: https://github.com/vim-syntastic/syntastic#faqpython

richey-v:
Closed.  I may or may not pursue fixing this upstream with pylint.

PR #2195, opened 2018-06-28
NASM output format

1995parham:
Add variable for configuring NASM output format. see [here](https://www.nasm.us/doc/nasmdoc7.html) for more information about NASM output formats.


lcd047:
This is not needed, the same can be accomplished by setting
```vim
let g:syntastic_nasm_nasm_args = '-f elf64'
```
Please consider reading the manual before working on patches.

Sorry, but no.

PR #2200, opened 2018-07-18
disable generate binary file when syntastic go file

elvis-macak:
Currently, each time when syntastic a go file using the default `go build mode`, it'll generate a binary file in the current folder,  which is annoying when using versioning tool like git.

Adding a `-o=/dev/null` will redirect it to /dev/null, more clear.

### DIFF
--- a/syntax_checkers/go/go.vim
+++ b/syntax_checkers/go/go.vim
@@ -60,7 +60,7 @@ function! SyntaxCheckers_go_go_GetLocList() dict
         let cleanup = 1
     else
         let cmd = 'build'
-        let opts = syntastic#util#bufVar(buf, 'go_go_build_args', s:go_new ? '-buildmode=default' : '')
+        let opts = syntastic#util#bufVar(buf, 'go_go_build_args', s:go_new ? '-buildmode=default -o=/dev/null' : '')
         let cleanup = 0
     endif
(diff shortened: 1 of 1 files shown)


lcd047:
Is option `-o` supported by all versions of `go` later than 1.4?  Was it actually tested with all these versions?  Does `-o=/dev/null` work on all OSes where syntastic and `go` can be installed?

If the answer to either question is no, then your patch is not needed.  The same effect can be achieved by setting `g:syntastic_go_go_post_args`.

elvis-macak:
- Is option -o supported by all versions of go later than 1.4?
   yes, it's from the official go tool, and supported at the beginning

- Was it actually tested with all these versions?
   It's a offical one from the strach, and I have test on go 1.4, and it works.

- Does -o=/dev/null work on all OSes where syntastic and go can be installed?
   **Unfortunately, it doesn't work in windows, so I also think this patch is not needed**


Finnaly I add option `let g:go_go_build_args = '-buildmode=default -o=/dev/null'` in my .vimrc (Unix like system), it can archive my requirement.

At last , thx @lcd047

lcd047:
> yes, it's from the official go tool, and supported at the beginning

If I recall correctly I tried this in the past, and it only worked in some cases, mainly because `-o` is supposed to point to a directory, or something like that.  Of course the real problem is `go` doesn't have a "check syntax" mode.  More points for dropping support for `go` in favor of [`vim-go`][vim-go].

PR #2203, opened 2018-08-01
mypy: ignore missing imports

pdxjohnny:
Importing stubless libraries fills up the bottom window with
module not found errors. `--ignore-missing-imports` flag
removes those messages allowing users to see issues
with their code.


lcd047:
So, what if I want warnings about missing modules?  Your patch prevents that from happening.

Also, unrelated: this is not needed.  It never was, from the first versions of syntastic.  You can achieve the same effect without a patch:
```vim
let g:syntastic_python_mypy_args = '--ignore-missing-imports'
```
Sorry, but no.

pdxjohnny:
Ok great, couldn't find that in the docs. Thanks

PR #2208, opened 2018-08-22
Adding remark-lint

pixelastic:
Hello,

I added a checker for parsing markdown through [remark-lint](https://github.com/remarkjs/remark-lint) (itself based on [remark](https://github.com/remarkjs/remark/tree/master/packages/remark-cli)).

It expects `remark` to be globally available [as a CLI tool](https://github.com/remarkjs/remark/tree/master/packages/remark-cli) and configured to use `remark-lint` by creating a `.remarkrc.js` similar to this one:
```javascript
exports.plugins = [
  'remark-preset-lint-recommended',
];
```

The `--no-stdout` prevents it from displaying the whole file in stdout. `--quiet` is supposed to only display warnings and errors, but I didn't see it having any impact when used with `--no-stdout`. I decided to keep it, just in case.

Output looks something like this:
image

I'm still no `errorformat` expert and I couldn't find a way to extract the file name (on its own line). So I cheated by setting a default `bufnr`. The checker surely has other rough edges; looking forward to see your patch on top of it to make it better @lcd047.

_Note that I checked the [Syntax Checker Guide](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide), mentioning that no new checker would be accepted until the API stabilized. But last update of this page is dating from 2016 so I assumed it has been stabilized since._

### DIFF
--- a/syntax_checkers/markdown/remark_lint.vim
+++ b/syntax_checkers/markdown/remark_lint.vim
@@ -24,7 +24,8 @@ function! SyntaxCheckers_markdown_remark_lint_GetLocList() dict
           \'args_before': '--quiet --no-stdout --no-color' })
 
     let errorformat =
-        \ '  %l:%c  %tarning  %m'
+        \ '%\s%#%l:%c%\s%#%tarning  %m  remark-lint,' .
+        \ '%\s%#%l:%c-%.%#%\s%#%tarning  %m  remark-lint'
 
     return SyntasticMake({
         \ 'makeprg': makeprg,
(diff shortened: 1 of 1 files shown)


lcd047:
I'm afraid I might not be able to look at this the next few days, sorry in advance for the delay.

pixelastic:
No problem, don't worry. I appreciate all the time you spent explaining the internals of syntastic on every PR I did in the past; take the time you need.

lcd047:
Plain `errorformat` can deal with filenames on separate lines (with `%D` / `%X`), but `%m` not matching till the end of line is problematic.  Syntastic has the preprocess hack to deal with this kind of limitations.

Merged, with this patch on top: c779dbb6.  Thank you!

PR #2213, opened 2018-10-03
In Ada, syntastic should check .ads files too

briot:
Hi,

It took me a while to realize that syntastic was not checking syntax in .ads files for Ada.
However, gcc is perfectly happy to handle those files, and it is convenient to have the
information immediately in vim.

Thanks
Emmanuel

### DIFF
--- a/syntax_checkers/ada/gcc.vim
+++ b/syntax_checkers/ada/gcc.vim
@@ -33,8 +33,7 @@ function! SyntaxCheckers_ada_gcc_GetLocList() dict
         \     '%f:%l:%c: %m,' .
         \     '%f:%l: %m',
         \ 'main_flags': '-c -x ada -gnats -gnatef',
-        \ 'header_flags': '-x ada -gnats -gnatef',
-        \ 'header_names': '\.ads$' })
+        \ 'header_flags': '-x ada -gnats -gnatef' })
 endfunction
 
 call g:SyntasticRegistry.CreateAndRegisterChecker({


lcd047:
`:h 'g:syntastic_ada_check_header'`

briot:
I could not find that variable in the code, after ending in c.vim.
Thanks, that's totally good enough, I'll close the pull request

briot:
In fact, I tried the variable and added various debug traces in syntastic.
After setting `g:syntastic_ada_check_header` to `1`, I see the check is run on the .ads file, as expected, but it gets passed an extra `-o /dev/null` parameter, which means the errors are not visible.

So on a .ads file, the command line is
   `gcc -x ada -gnats -gnatef -c -o /dev/null foo.ads`
but on a .adb the command line is
   `gcc -c -x ada -gnats -gnatef foo.adb`

The former command outputs `gnat1: incorrect object file extension`, while the second properly shows errors. Both cases set the exit status at 1.

lcd047:
How about:
```sh
gcc -S -x ada -gnats -gnatef -c -o /dev/null foo.ads
```
Would that work?

briot:
Yes, that works indeed:

```
> gcc -S -x ada -gnats -gnatef -c -o /dev/null  General/Logging/loggers_builtins.ads
General/Logging/loggers_builtins.ads:39:13: subtype indication expected
> gcc -x ada -gnats -gnatef -c -o /dev/null  General/Logging/loggers_builtins.ads
gnat1: incorrect object file extension
```

lcd047:
Ok then, commit df9f7057.  Thank you!

briot:
That was extremely fast, for a language that I suspect is not your core business ! :-)

Thanks again

PR #2214, opened 2018-10-14
Update README.markdown

arten1337:


lcd047:
Err, no.  The TOC is fine just the way it is. :smile:

arten1337:
okay

PR #2215, opened 2018-10-17
checkpatch: Catch strict warnings

SoldierDash:
When the "--strict" or "--subjective" options are added to checkpatch's
arguments, it will return additional warnings of this format type:

> file.c:123: CHECK: Alignment should match open parenthesis

The extra warnings can be enabled with:

> let g:syntastic_c_checkpatch_args = "--strict"

Signed-off-by: Brandon Maier <brandon.maier@rockwellcollins.com>


lcd047:
Merged, thank you.

PR #2216, opened 2018-10-17
checkpatch: Enable tree checking if available

SoldierDash:
If checkpatch has access to a Linux source tree, it can find additional
warnings. Add a new option to let users manually enable/disable the
'--no-tree' option. If no selection is made, attempt to autodetect tree
support.

Signed-off-by: Brandon Maier <brandon.maier@rockwellcollins.com>

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -579,6 +579,15 @@ Checker options~
 This checker is initialised using the "makeprgBuild()" function and thus it
 accepts the standard options described at |syntastic-config-makeprg|.
 
+Additionally:
+
+                                            *'g:syntastic_c_checkpatch_tree'*
+Type: boolean
+Default: unset
+When set to '0', will call checkpatch with '--no-tree' which disables warnings
+that require the full Linux source tree to run. If unset, Syntastic will
+attempt to autodetect if tree is available.
+
 ------------------------------------------------------------------------------
 3. ClangCheck                                        *syntastic-c-clang_check*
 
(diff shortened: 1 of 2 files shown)


lcd047:
The same effect can be achieved with the existing options, without any patch.  Sorry, but no.

SoldierDash:
Ah, I hadn't realized it as it's not documented in Checkpatch. For reference, setting `g:syntastic_c_checkpatch_post_args = '--tree' ` will override '--no-tree', as Checkpatch  respects the last option passed.

However this still makes the default '--no-tree', which I don't think is user friendly. Would you be more partial if I got rid of the configurable option and system() call, and made the default so that if the detected checkpatch exec is "./script/checkpatch.pl" or "./scripts/checkpatch" it removes '--no-tree'. That way users get '--tree' when they're running from the Linux tree root.

```
 if executable('checkpatch')                         
     let g:syntastic_c_checkpatch_exec = 'checkpatch'
 elseif executable('./scripts/checkpatch.pl')                                         
     let g:syntastic_c_checkpatch_exec = fnamemodify('./scripts/checkpatch.pl', ':p')
+    let g:syntastic_c_checkpatch_tree = 1
 elseif executable('./scripts/checkpatch')                                            
     let g:syntastic_c_checkpatch_exec = fnamemodify('./scripts/checkpatch', ':p')    
+    let g:syntastic_c_checkpatch_tree = 1
 endif               
```

lcd047:
Sorry, patches that modify defaults are accepted only if there are very strong reasons for the change (as in security risks or similar).  You don't want a developer tool that behaves slightly different every day, do you.

PR #2218, opened 2018-10-20
Implemented basic Zig syntax checker.

DavidYKay:
Implemented a very simple checker for the [Zig language].

Implementation currently invokes "Zig Build" and thus depends on using a Zig buildfile.

Does not yet support GetHighlightRegex.

### DIFF
--- a/plugin/syntastic/registry.vim
+++ b/plugin/syntastic/registry.vim
@@ -114,6 +114,7 @@ let s:_DEFAULT_CHECKERS = {
         \ 'yang':          ['pyang'],
         \ 'yara':          ['yarac'],
         \ 'z80':           ['z80syntaxchecker'],
+        \ 'zig':           ['zig'],
         \ 'zpt':           ['zptlint'],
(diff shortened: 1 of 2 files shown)


lcd047:
The Zig language seems to be experimental.  Please consider turning your patch into an [external checker][guide].  This is the intended use for external checkers:  the syntastic checker should be distributed with some kind of package that adds support for Zig files to Vim.  This way the checker can be maintained by people who actually use the language and care about it.

[guide]: https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#external

DavidYKay:
OK, I ended up switching to ALE for my own personal use. Happy to move this to an external checker if someone else pipes in and wants this merged into Syntastic. Otherwise, I'll most likely just leave this PR as-is.

Thanks much for your reply and your work on Syntastic! :smile:

lcd047:
Absolutely, ALE is the way to go.

PR #2220, opened 2018-10-26
Ignore warnings in javascript/eslint IsAvailable() check

ccraciun:
Resolves: https://github.com/vim-syntastic/syntastic/issues/2219


lcd047:
The checker for `eslint` is supposed to be able to use other executables than `eslint` itself, in order to accommodate the zillion of linters derived from `eslint`.  Your patch breaks that assumption.  Sorry, but no.

PR #2227, opened 2018-11-01
Checker python/mypy: allow return code 2

ahnurmi:
Allow mypy checker to return status code 2 on Python syntax error.

The issue is encountered if a previous checker does not catch the syntax error.


lcd047:
I disagree.  A syntax error in the input leads to an abnormal termination of `mypy`, there is no point in trying to parse its output in that case.

ahnurmi:
Ok! Very likely I’m missing something obvious, but `mypy` does print the error line, e.g.`foo.py:108: error: invalid syntax`, which gets parsed fine.

PR #2229, opened 2018-11-06
Checker ansible/ansible_lint: fix errorformat

netzvieh:
ansible-lint now has a slightly different outputformat for errors:

```
prepare-host.yaml:24: [E302] chmod used in place of argument mode to
file module
```

Append the errorformat so this gets recognized by syntastic.


lcd047:
Merged, thank you!

Issue #1948, opened 2017-01-20
Puppet parser validate not reporting for 'Syntax error at end of file'


Joshua-Snapp:
I noticed if I left a closing brace off at the end of a standard Puppet manifest .pp file, Syntastic would fail to report the error generated by the `puppet parser validate` command.

Here's the super simple Puppet manifest code that should reproduce what I'm reporting. I just threw that bit of code in a test.pp file.

```
class test {
```

I enabled Syntastic debugging.
`let g:syntastic_debug = 3`

Here's the Syntastic debug info found in :messages.

```
"test.pp" [New] 1L, 13C written
syntastic: 74.586135: g:syntastic_version = '3.8.0-14 (Vim 800, Linux)'
syntastic: 74.586246: &shell = '/bin/bash', &shellcmdflag = '-c', &shellpipe = '2>&1| tee', &shellquote = '', &shellredir = '>%s 2>&1', &shelltemp = 1, &shellxquote = '
', &autochdir = 0, &shellxescape = ''
syntastic: 74.586541: UpdateErrors (auto): default checkers
syntastic: 74.586781: CacheErrors: default checkers
syntastic: 74.587269: g:syntastic_aggregate_errors = 0
syntastic: 74.587418: getcwd() = '/home/jksnapp/itecs/puppetcode/forks/realmlinux-daniel'
syntastic: 74.588668: CacheErrors: Invoking checker: puppet/puppet
syntastic: 74.899402: system: command run in 0.310503s
syntastic: 74.899965: SyntasticMake: called with options: {'errorformat': '%-Gerr: Try ''puppet help parser validate'' for usage,%-GError: Try ''puppet help parser vali
date'' for usage,%A%t%*[a-zA-Z]: %m at %f:%l:%c,%A%t%*[a-zA-Z]: %m at %f:%l', 'makeprg': 'puppet parser validate --color=false test.pp'}
syntastic: 75.422439: system: command run in 0.522203s
syntastic: 75.422596: checker output: ['Error: Could not parse for environment production: Syntax error at end of file  in /home/jksnapp/itecs/puppetcode/forks/realmlin
ux-daniel/test.pp', '']
syntastic: 75.422918: raw loclist: [{'lnum': 0, 'bufnr': 0, 'col': 0, 'valid': 0, 'vcol': 0, 'nr': -1, 'type': '', 'pattern': '', 'text': 'Error: Could not parse for en
[...]

PR #2239, opened 2019-01-06
Puppet: caught error output w/o line and column

Slm0n87:
solves #1948


lcd047:
Sadly, this breaks error sorting (cf. `:h syntastic-config-sort`).

Slm0n87:
thx for the quick feedback. you are right.  
in my case the output of the puppet-parser is shown at first - before the output of puppet-lint. which is absolut OK for me from usibility perspective, but I understand that you prefer to assign the error to a line :(

PR #2241, opened 2019-01-21
Add validator.nu checkers for XHTML and SVG

kevinoid:
The [validator.nu](https://validator.nu) service supports XHTML and SVG in addition to HTML.  This pull request refactors html/validator so that the API calling logic can be shared, then adds checkers for XHTML and SVG by calling the shared function.  Let me know what you think.

Thanks for considering,
Kevin

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -107,6 +107,7 @@ SYNTAX CHECKERS BY LANGUAGE                          *syntastic-checkers-lang*
     Solidity.................................|syntastic-checkers-solidity|
     SQL......................................|syntastic-checkers-sql|
     Stylus...................................|syntastic-checkers-stylus|
+    SVG......................................|syntastic-checkers-svg|
 
     Tcl......................................|syntastic-checkers-tcl|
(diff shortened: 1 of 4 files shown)


lcd047:
I like to change things in a [single place][dups], so please see f6ec8a41 instead.  The interface is horrendous (mainly because the API of `validator.nu` is horrendous), but syntastic is practically dead anyway, so maybe that's good enough.

[dups]: https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#duplicates

kevinoid:
That works for me.  Thanks @lcd047!

> syntastic is practically dead anyway

That's news to me.  Thanks for mentioning it.  Could you elaborate?

lcd047:
People have been asking for asynchronous checking for a long time.  Syntastic is unlikely to get asynchronous checking, mainly because of my health condition, and my vanishing interest in Vim development.  So many former syntastic users have migrated to [ALE][ale].

[ale]: https://github.com/w0rp/ale

kevinoid:
Good to know.  Sorry to hear about your health condition.  Thanks again for adding SVG+XHTML and for the information!

PR #2244, opened 2019-01-26
Checker php/phpstan: fix errorformat option (BC break in PHPStan 0.11)

jwmwalrus:
PHPStan version 0.11 changes the command-line flag from `--errorFormat` to `--error-format`.  That change makes the checker unusable.

This PR fixes it.


lcd047:
Merged, thank you.

PR #2245, opened 2019-01-28
Correct syntax error in oclint.vim

JackStillwell:
Path: syntastic/syntax_checkers/c/oclint.vim
Line: 28
Char: 106
Correction: ' ' -> ')'


lcd047:
Merged, thank you.

PR #2246, opened 2019-01-28
Checker html/w3: Add support for SVG and XHTML

kevinoid:
This PR applies the same changes that were applied to `html/validator` in vim-syntastic/syntastic#2241 to `w3/html` so that it can be used to validate SVG and XHTML as well.  It makes `html/w3` nearly identical to `html/validator`, except for `errorformat`, and for `w3` lacking some of `validator`'s options.

Thanks for considering,
Kevin

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -3175,7 +3175,7 @@ details:
     https://validator.w3.org/source/
 
 As a syntastic linter, you can validate your files against the online service
-(see http://validator.w3.org/), or you can install it from sources and run it
+(see https://validator.w3.org/), or you can install it from sources and run it
 as a local service:
 
     https://github.com/w3c/markup-validator/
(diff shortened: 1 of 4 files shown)


lcd047:
I'm afraid I don't understand what are you trying to achieve here.  Your patch is either too much or too little to check SVG files.  If you're trying to duplicate the functionality of https://validator.w3.org/ just setting `Content-Type` is not going to cut it, and if there is no documented way to distinguish between, say, SVG 1.0 and SVG 1.1 with the API.  On the other hand, if you're trying to do DTD-based validation you need to supply the relevant DTDs, and your patch doesn't provide the means to do that.

Unrelated: a feature that isn't documented can be used (and maintained) only by its author. _shrug_

kevinoid:
> I'm afraid I don't understand what are you trying to achieve here. Your patch is either too much or too little to check SVG files. If you're trying to duplicate the functionality of https://validator.w3.org/ just setting Content-Type is not going to cut it.

You are right.  Good catch.  I expected it to detect SVG without a DTD in the same way it detects HTML and XHTML without a DTD, but currently it validates SVG without a DTD as generic XML.

> There seems to be no documented way to distinguish between, say, SVG 1.0 and SVG 1.1 with the API.  On the other hand, if you're trying to do DTD-based validation you need to supply the relevant DTDs, and your patch doesn't provide the means to do that.

You can do this by setting the `doctype` POST parameter to `SVG 1.0` or `SVG 1.1` (or `SVG 1.1 Tiny`, `SVG 1.1 Basic`).  The validator will add the DTD if it is missing and the `doctype` parameter is specified.

I could add code to detect the version from the `version=` attribute on the `<svg>` element.  This would probably be painful to do correctly without an external XML parser.  Alternatively, I could set `doctype` from `g:syntastic_{type}_w3_doctype` (and `b:syntastic_{type}_w3_doctype`?) with a default of `SVG 1.1` for SVG.  Or something else?  Do you have a preference for how it should work?

> Unrelated: a feature that isn't documented can be used (and maintained) only by its author. shrug

Ha.  That was an oversight on my part.  Good catch.  Docs updated.

lcd047:
> I could add code to detect the version from the `version=` attribute on the `<svg>` element.  This would probably be painful to do correctly without an external XML parser.  Alternatively, I could set `doctype` from `g:syntastic_{type}_w3_doctype` (and `b:syntastic_{type}_w3_doctype`?) with a default of `SVG 1.1` for SVG.  Or something else?  Do you have a preference for how it should work?

Leaving this to the user (i.e. reading `doctype` from a variable) is probably the only reasonable approach.  Reading it from the `version` attribute doesn't allow user to override it (in case somebody wants to check backwards compatibility, or whatever), and parsing XML in Vim is a hack at best.

kevinoid:
I agree.  PR updated.  Let me know what you think.

lcd047:
Only in the new checkers `svg/w3` and `xhtml/w3`.

kevinoid:
Gotcha.  That makes sense to me.  Thanks!  PR updated.

lcd047:
Thank you.  Merged with this patch on top: 8c76c04.

kevinoid:
Great!  Thanks for all of the feedback and for merging it @lcd047!

PR #2249, opened 2019-03-06
Fix typo in eslint syntax_checker

AnthoniM:
When defining `makeprg` with `g:syntactic_javascrit_eslint_generic=0`, the added `compact` option is missing the `--` prefix.

### DIFF
--- a/syntax_checkers/javascript/eslint.vim
+++ b/syntax_checkers/javascript/eslint.vim
@@ -42,7 +42,7 @@ function! SyntaxCheckers_javascript_eslint_GetLocList() dict
             \ "'--config ' . syntastic#util#shexpand(OLD_VAR)")
     endif
 
-    let makeprg = self.makeprgBuild({ 'args_before': (g:syntastic_javascript_eslint_generic ? '' : '-f compact') })
+    let makeprg = self.makeprgBuild({ 'args_before': (g:syntastic_javascript_eslint_generic ? '' : '-f --compact') })
 
     let errorformat =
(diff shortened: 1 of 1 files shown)


lcd047:
`compact` is actually meant to be the argument to `-f`, not a separate option.  Sorry, but no.

AnthoniM:
The syntax_checker javascript/eslint.vim runs the following command upon saving a file app.js : `npm run lint -f compact app.js`. Form the error log, it seems somehow that npm sees that `-f` option as the `--force` option and thus expect the compact option to be also a file. I agree that my fix what no a fix and thank you for the precision, but why not use `--format=compact` instead of `-f compact` ? How is the `-f` option not supposed to be interpreted as the force option by npm ? Is there something I am missing ?

This the outcome of `npm run lint -f compact app.js` :
>npm WARN using --force I sure hope you know what you are doing.
>
>library@1.0.0 lint C:\Users\anthoni.manseau\Langages\Javascript\NodeJSExpress\library
>eslint app.js "compact"
>
>Oops! Something went wrong! :(
>
>ESLint: 5.15.1.
>No files matching the pattern "compact" were found.
>Please check for typing mistakes in the pattern.

lcd047:
> The syntax_checker javascript/eslint.vim runs the following command upon saving a file app.js : `npm run lint -f compact app.js`.

Nope, syntastic doesn't run that by default.  It does it only if you mess with the `exe` and `exec` variables, in which case it's up to you to make sure what you're doing makes sense.

PR #2250, opened 2019-03-07
Checker haskell/hlint: supress colors

phadej:


lcd047:

phadej:
`hlint` might have configuration which forces colored output, even when terminal doesn't support it. This makes `hlint` not emit color codes.

I don't know if it's better to strip ANSI color codes, but I don't know how to do that.

lcd047:
The downside to this change is that option `--color` is supported only since hlint version v1.9.4.  Your patch needs to take that into account.

phadej:
The command line parsing library which `hlint` uses, got support for `--numeric-version` in

```
0.10.8, released 2014-07-21
    Avoid compilation warnings on GHC 7.8
    #9, add --numeric-version flag
```

It's unfortunately not tied to `hlint`'s version.

The `--version` flag outputs something like

```
HLint v2.1.15, (C) Neil Mitchell 2006-2019
```

which I don't know how to parse with `vimscript`

lcd047:
Ok, please try 1089ee52.

phadej:
@lcd047 that works, thanks!

PR #2257, opened 2019-05-05
Adds warnings support for the solidity/solc checker.

ajsantander:
Warnings such as:
```
SimpleStorage.sol:7:16: Warning: Unused function parameter. Remove or comment out the variable name to silence this warning.
  function set(uint x) public {...
```
are not detected by the solc.vim checker. This adds a new format to the local errorformat variable so that they are.


lcd047:
Merged, thank you.

PR #2264, opened 2019-05-28
Fix possible spelling mistake

Toothless204:


lcd047:
It isn't a typo.  Next time please consider fixing only actual problems. :smile:

PR #2266, opened 2019-06-06
Scilla syntax checker

bogdan:
Scilla is a programming language of Zilliqa blockchain

https://scilla-lang.org/

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -100,6 +100,7 @@ SYNTAX CHECKERS BY LANGUAGE                          *syntastic-checkers-lang*
 
     SASS.....................................|syntastic-checkers-sass|
     Scala....................................|syntastic-checkers-scala|
+    Scilla...................................|syntastic-checkers-scilla|
     SCSS.....................................|syntastic-checkers-scss|
     Sh.......................................|syntastic-checkers-sh|
     Slim.....................................|syntastic-checkers-slim|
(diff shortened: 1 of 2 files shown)


lcd047:
Scilla isn't something I can check easily, and it seems pretty specialized.  Consequently this checker is probably best fit for staying [ext][external].

A couple of suggestions though:
- you shouldn't add `exe` to `makeprgBuild()`, setting `exec` in `CreateAndRegisterChecker()` takes care of  all relevant details
- you don't need an ad-hoc variable `syntastic_scillachecker_stdlib_path`, the standard mechanism of `syntastic_scilla_scillachecker_args` is built in and much more flexible.

[ext]: https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#external

bogdan:
Ok, I applied all your suggestions. You may or may not merge it... that is fine. I just got it online if someone would need it and can quickly copypaste without knowing all syntastic details.

It would be easier if I can make such generic checkers in single line like:

``` vim
call g:SyntasticRegistry.CreateAndRegisterChecker({
    \ 'filetype': 'scilla',
    \ 'name': 'scillachecker',
    \ 'exec': 'scilla-checker',
    \ 'errorformat': '%f:%l:%c:%m'})
```

That would make me feel like it is not worth inventing a vim plugin.

lcd047:
It's possible, provided you're willing to give up some functionality: `:h write-compiler-plugin`.

bogdan:
Well, I don't want to give up functionality obviously... I just think that my code can be 5x shorter if `CreateAndRegisterChecker` would support `errorformat` option. I think many syntax checkers can be 1 line of code if `CreateAndRegisterChecker` would support all the options of `makergBuild` and `SyntasticMake`.

lcd047:
Since you insist of turning this into an issue: I just run a quick `grep` through the code.  Currently, out of 187 checkers that use `makeprgBuild()`, 40 are calling `makeprgBuild({})`.  Out of these, 9 also have a trivial `SyntasticMake()` and nothing in between, and would benefit from your suggestion.  Fortunately syntastic design wasn't optimized for your preferences.

Unrelated: as you might have noticed, syntastic is all but dead.  People seem to use [ALE][ale] these days.

[ale]: https://github.com/w0rp/ale

PR #2267, opened 2019-06-15
Loclist follow

elbeardmorez:
For your consideration. 

As per the first commit message..

-adds 'SyntasticLoclist.nearest()' which sets the location list
position to the item nearest the current buffer's cursor position. use
of the 'CursorMoved' hook allowing this to be used to 'sync' your
cursor with the location list
-enabled / disabled via 'g:syntastic_loc_list_follow' int

Notes:
1. Touches vimscript only.
2. I've had a simpler version of this working a while now, the latter tweaks/commits around interline movement are new, but appear to be working well.
3. Please be aware my vimscript foo isn't great so elementary vimscript faux-pas should be looked for!
4. The attempt at integration feels a little flaky to me too. Although I believe I followed an existing pattern, using a buffer variable as 'the switch' doesn't seem ideal, and also finding I had to move the hook in 'SyntasticRefreshCursor' to avoid short-circuits for 'private line' functionality felt hairy.
5. There may be a better (built-in) mechanism to arrive at the current/last item too (e.g. it's surely calculated somewhere given it's displayed in the cmd window (albeit only when the cursor is actually on the item)) ..but alas, the functionality I wanted, based on my own preference, was 'whatever's nearest', full stop, and that's what's implemented.
6. I think the use of the buffer var storing last position, in order to quickly get one to 'usually within the near vicinity of' the correct line on each iteration is a good enough approach to scale with silly large list sets - I see no lag at all.. ..but then maybe my setup is simplistic (few plugins) compared to others and thus this might require consideration. As noted in the last commit, a short-circuit was necessarily removed to get the interline aspect working ..so maybe that could be optional.

My sincere thanks for this plugin, it's just fantastic.

### DIFF
--- a/doc/syntastic.txt
+++ b/doc/syntastic.txt
@@ -499,6 +499,14 @@ When set to 3 the error window will be automatically opened when errors are
 detected, but not closed automatically. >
     let g:syntastic_auto_loc_list = 3
 <
+                                                 *'syntastic_loc_list_follow'*
+Type: integer
+Default: 0
+Use this option to enable automatic updating of the selected location list
+item based on the current cursor position. The nearest item with be selected.
+>
+    let g:syntastic_loc_list_follow = 1
+<
                                                  *'syntastic_loc_list_height'*
 Type: integer
 Default: 10
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -101,6 +101,7 @@ let g:_SYNTASTIC_DEFAULTS = {
         \ 'id_checkers':              1,
         \ 'ignore_extensions':        '\c\v^([gx]?z|lzma|bz2)$',
         \ 'ignore_files':             [],
+        \ 'loc_list_follow':          0,
         \ 'loc_list_height':          10,
         \ 'nested_autocommands':      0,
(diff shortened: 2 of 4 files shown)


lcd047:
Sorry to disappoint you, but this seems to be outside the scope of syntastic.  The purpose of syntastic is to run linters, and build `loclist`s based on their output; navigating said `loclist`s is intentionally left to the user.  Your patch could be a stand-alone plugin though: if I understand it correctly it doesn't have anything specific to syntastic, it just helps with with navigation through `loclist`s and `qflist`s.

Unrelated: syntastic is essentially dead these days, people seem to use [ALE][ALE] instead these days.

[ALE]: https://github.com/w0rp/ale

elbeardmorez:
No probs. I saw it as extending what 'show' and 'jump' features currently do but understand every project has its lines. I'll now have equal fun ripping it back out and creating a standalone plugin 🤣

PS. Thanks for the heads up. 👍

elbeardmorez:
In case anyone stumbles upon this effort with enough interest, see [vim-loclist-follow](https://github.com/elbeardmorez/vim-loclist-follow)

PR #2271, opened 2019-06-26
Add a new Go linter: gopls

jpguerard:
gopls is a new Language Server Protocol implementation for Go. With recent releases, it can also be used as a standalone linter.

This new checker uses gopls to check Go programs.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -2516,8 +2516,9 @@ The following checkers are available for Go (filetype "go"):
     3. GolangCI-Lint............|syntastic-go-golangci_lint|
     4. Golint...................|syntastic-go-golint|
     5. Go Meta Linter...........|syntastic-go-gometalinter|
-    6. gotype...................|syntastic-go-gotype|
-    7. vet......................|syntastic-go-govet|
+    6. gopls....................|syntastic-go-gopls|
+    7. gotype...................|syntastic-go-gotype|
+    8. vet......................|syntastic-go-govet|
 
 ------------------------------------------------------------------------------
 1. go                                                        *syntastic-go-go*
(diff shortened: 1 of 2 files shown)


lcd047:
Your commit doesn't include the actual checker. :smile: 

Anyway, please consider adding this to [ALE][ALE] and / or [vim-go][vim-go] instead.  The `go` checkers in syntastic are still around only for backwards compatibility, they don't serve any real purpose these days.

[ALE]: https://github.com/w0rp/ale

jpguerard:
From my end, the checker seems to be in the commit.
gopls is already supported by ALE and vim-go.
I'm using syntastic, so this checker was useful to me.

lcd047:
> From my end, the checker seems to be in the commit.

Right, I can see it now too.  So either I discovered a race condition in GitHub, or I was just imagining things.  I guess I'll spend the rest of my life wondering.

> I'm using syntastic, so this checker was useful to me.

Well, the point still stands: `go` checkers should have been deleted a long time ago, and adding _more_ checkers for me to take care of is not a step towards that direction.  Also, syntastic is essentially dead these days.

Anyway, about the checker:

* In `IsAvailable()` you should probably parse the output rather than check `v:shell_error`.  The latter is known not to work reliably under the `cmd.exe` shell on Windows, so it should be avoided when possible.
* You probably want `let errorformat = ... '%f:%l:%c-%\d%\+: %m,' ...`.  Backslashes have no special meaning inside single quotes.
* The `exec` in `CreateAndRegisterChecker()` is redundant but not wrong.

jpguerard:
* Parsing the output in IsAvailable would no really be helpful. The first versions of gopls did not understand the commands "version" and "check" (this is the case of the gopls available in Debian). So either "gopls version" fails as it does not understand "go version", and we can't use it. Or it succeed, and can be used.
* I've correcter errorformat.

jpguerard:
I also tested gopls on Windows. It does correctly return an error code if it does not understand the option provided.

Issue #1957, opened 2017-02-05
shellcheck should have the -x option


dunatotatos:
When sourcing an other file, shellcheck has the ability to follow the sourced file thanks to the directive
`# shellcheck source=/path/to/file`
To follow the source, shellcheck must be used with the `-x` option, or the warning [SC1091](https://github.com/koalaman/shellcheck/wiki/SC1091) is triggered.

Is it possible to add this option by default? Or provide an easy way to enable the option?

lcd047:
Please read `:h syntastic-sh-shellcheck` and `:h syntastic-config-makeprg`.

dunatotatos:
Thank you for the doc.
Would it be a good thing to have this option by default?
If yes, I can provide a pull request.

lcd047:
> Would it be a good thing to have this option by default?

I suppose that would be a good question for the author of `shellcheck`.  As far as syntastic is concerned, defaults should change only for very strong reasons.  Otherwise somebody else would have to change his setup because it doesn't match yours. _shrug_

iNode:
Just in case if somebody will need this in the future.

Option to enable this will be `let g:syntastic_sh_shellcheck_args="-x"`

qualitymake:
> let g:syntastic_sh_shellcheck_args="-x"

How to change this ? please give me the steps

fourdollars:
> > let g:syntastic_sh_shellcheck_args="-x"
> 
> How to change this ? please give me the steps

Edit your ~/.vimrc and add the line of `let g:syntastic_sh_shellcheck_args="-x"`
I just confirmed that It works.

PR #2277, opened 2019-07-30
Checker shellcheck: check external sources by default.

fourdollars:
-x, --external-sources
              Follow 'source' statements even when the file is not specified as input.  By default, shellcheck will only follow files specified on the command line (plus /dev/null).  This option allows following any file  the script may source.


lcd047:
Please see #1957.  Also, the patch is wrong: if you add `-x` to `args_after` it can't be overridden by the users.

Sorry, but no.

PR #2278, opened 2019-08-03
Added note about installing the lintr package for r/rmd lintr support

khughitt:
Added a note about the need to install the "lintr" package for lintr to work.. Otherwise there are no error messages / warnings, but 'lintr' won't show show up in the list of available linters.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -5426,6 +5426,15 @@ GitHub for details:
 
     https://github.com/jimhester/lintr
 
+Installation~
+
+In order to use this linter, you must first install the "lintr" R package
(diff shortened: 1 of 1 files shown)


lcd047:
Why document the obvious though?  `lintr` is an R module, why would anybody expect the corresponding syntastic checker to be available when the module isn't installed?

khughitt:
Ha. okay fair point. It took me a while to figure this out, but that's only because I didn't take the time to first read the manual and assumed (incorrectly) that Syntastic had at least some support for linting itself. I see now that it is really just wrapping a bunch of other language-specific linters which, in hindsight, makes a lot more sense.

Closing the PR.

PR #2302, opened 2020-04-06
Update README.markdown

raywill:
:Helptags => :help tags


lcd047:
Nope, sorry.  The point is to compile help tags, not to read help about tags.

PR #2306, opened 2020-04-22
Add install instructions for vim 8+

solderjs:
Add simple instructions for using syntastic with native vim 8 packages.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -127,6 +127,23 @@ checkers.
 
 <a name="installpathogen"></a>
 
+### 2.1.1\. Installing syntastic with Vim 8+
+
+Add `packloadall` to your `.vimrc` (if you haven't already):
+
+`.vimrc`:
+
+```vim
+packloadall
+```
+
+Clone `syntastic` into your `~/.vim/pack/plugins/start` directory:
+
+```bash
+mkdir -p ~/.vim/pack/plugins/start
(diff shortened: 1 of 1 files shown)


lcd047:
What problem does your patch solve?  According to the README file:

> Starting with Vim version 7.4.1486 you can also load syntastic using the standard mechanism of packages, without the help of third-party plugin managers (see `:help packages` in Vim for details).

This seems to be clear enough, nobody has ever expressed any confusion about it.

solderjs:
Now that macOS ships with Vim that supports plugins, it makes to me to make the standard way more prominent (and perhaps make the old way less prominent).

I never saw that notice before (buried in lots of text) - and it's not clear to me.

Just trying to make it as clear how to install this the new standard way as it is clear how to install the old way with pathogen.

lcd047:
There isn't anything special about syntastic, it can be installed the same way as any other Vim plugin.  Syntastic has now been obsoleted by [ALE][ALE] (which can do asynchronous checking and interface with language servers), so it's only real use these days is for situations where you need to run it under Vim 7.  Hence the reference to Pathogen, since Vim 7 added a package manager only in the late stages.

TL;DR: there is little point in grooming a dead horse.

solderjs:
I've never heard of Ale. If that's the replacement... should not that go in the README as well?

PR #2308, opened 2020-05-02
Avoid calling `uname` whenever possible

telotortium:
Calling `uname` every time Vim starts takes a noticeable time, so avoid calling
it whenever possible. This reduces the running time of `plugin/syntastic.vim`
from ~25 ms to ~5 ms on my machine.

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -52,8 +52,21 @@ if !exists('g:syntastic_shell')
     let g:syntastic_shell = &shell
 endif
 
-if s:_running_windows
+" Determine `uname` without running the actual command, for faster startup.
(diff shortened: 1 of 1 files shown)


lcd047:
The point of running `uname` is to check that Vim and syntastic are (still) configured correctly, and can (still) run external programs.  The result of `uname` is used exclusively for debug logs, syntastic itself doesn't care about anything of that kind.

Anyway, if you're trying to reduce startup time with 20 ms you aren't just barking up the wrong tree, you're in the wrong forest altogether.  Please consider using [ALE][ALE] instead: unlike syntastic it can do asynchronous checks, and it's actively maintained.

PR #2317, opened 2020-06-19
Update possible mypy return codes, update documentation

everybodyeverybody:
Added an extra section to the documentation which demonstrates how one
can install syntastic from git using vim packages.

mypy will return 2 as a potential error code in cases of improperly
formatted python syntax. 

For example:
```python
#!/usr/bin/env python
a =
```
will provide a return code of 2 for invalid syntax. 

This updates the mypy configuration to accept this return code.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -21,7 +21,8 @@
 1. [Introduction](#introduction)  
 2. [Installation](#installation)  
 2.1. [Requirements](#requirements)  
-2.2. [Installing syntastic with Pathogen](#installpathogen)  
+2.2. [Installing syntastic with Pathogen](#installpathogen)
+2.3. [Installing syntatic with vim8 packages](#installvim8)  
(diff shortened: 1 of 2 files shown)


lcd047:
Your patch does two unrelated things.  First, it allows exit code 2 for `mypy`.  This is undesirable because exit 2 corresponds to [crash, bad arguments, and other non-standard conditions](https://github.com/python/mypy/issues/6003).  Of these only syntax error produces a meaningful error message for syntastic to parse, meaning that with your patch conditions such as "No such file or directory" get silently ignored.  You really do want to know about these.

Second, it adds an example of installing syntastic "directly into vim version 8+ from git".  This is too restrictive to be useful, redundant since the main README already mentions `:help packages`, and struggles to get the Markdown syntax right.

Sorry, but no.

PR #2326, opened 2020-07-26
check vue files with stylelint

oprudkyi:
add ability to use stylelint to check *.vue files

### DIFF
--- a/syntax_checkers/css/stylelint.vim
+++ b/syntax_checkers/css/stylelint.vim
@@ -22,6 +22,7 @@ set cpo&vim
 let s:args_after = {
     \ 'css':  '-f json',
     \ 'html': '-f json',
+    \ 'vue':  '-f json',
     \ 'less': '-f json -s less',
     \ 'scss': '-f json -s scss' }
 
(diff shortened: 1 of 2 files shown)


lcd047:
This isn't really needed, you can run `stylelint` against any file with `:SyntasticCheck css/stylelint`.  Syntastic currently doesn't have any `vue`-specific checkers, and there's no point pretending `stylelint` can play that role.  Sorry, but no.

On a side note, syntastic is essentially dead.  You might want to take a look at [ALE][ALE] instead.

oprudkyi:
Hi @lcd047 

vue file is a bundle of html/js/css/scss https://vuejs.org/v2/guide/single-file-components.html and having stylelint as one of checkers is essential for me. but ok, understood. anyway thank you for useful plugin

lcd047:
Please see `:h 'syntastic_filetype_map'`.

oprudkyi:
yes, I have tried it, but without success
- stylelint breaks without
```
    \ 'vue':  '-f json',
``` 
in syntax_checkers/css/stylelint.vim
- I need both checkers, eslint and stylelint to run over vue files, and syntastic_filetype_map didn't work for me correctly (I don't remember exactly, but it seems like it allowed to use single checker only)

lcd047:
I see.  Ok, merged with this patch on top: 83837a6e.

PR #2328, opened 2020-08-13
Fix issues in Objective-C++ gcc checker

dpoggi:
* `-std=gnu99` has been changed to `-std=gnu++14` in the default compiler options to reflect the default in the current stable version of Xcode (`-std=gnu99` previously broke the checker)
* `SyntaxCheckers_objcpp_gcc_IsAvailable()` now checks for the existence of the correct global when deciding whether to use the default compiler (previously checked for the C compiler)
* Default compiler executables have been changed to reflect that Objective-C++ is a flavor/dialect of C++, not C

### DIFF
--- a/syntax_checkers/objcpp/gcc.vim
+++ b/syntax_checkers/objcpp/gcc.vim
@@ -16,15 +16,15 @@ endif
 let g:loaded_syntastic_objcpp_gcc_checker = 1
 
 if !exists('g:syntastic_objcpp_compiler_options')
-    let g:syntastic_objcpp_compiler_options = '-std=gnu99'
+    let g:syntastic_objcpp_compiler_options = '-std=gnu++14'
 endif
 
 let s:save_cpo = &cpo
 set cpo&vim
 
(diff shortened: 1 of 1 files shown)


sanphamdoimenhphattai:
https://event.onlinecoachinggroup.com/event/webinar/5f35fc5269a40[](url)

lcd047:
> * `-std=gnu99` has been changed to `-std=gnu++14` in the default compiler options to reflect the default in the current stable version of Xcode (`-std=gnu99` previously broke the checker)

Changes to the defaults are accepted only if they address security issues or similar, sorry.  The defaults are there to be overridden, and you need to set compiler flags to match your project anyway.

> * `SyntaxCheckers_objcpp_gcc_IsAvailable()` now checks for the existence of the correct global when deciding whether to use the default compiler (previously checked for the C compiler)

Right, please see 25b816d9.

> * Default compiler executables have been changed to reflect that Objective-C++ is a flavor/dialect of C++, not C

What problem does this solve?  Syntastic runs `gcc -x <language>` rather than distinguish between `gcc` and `g++`.

PR #2333, opened 2020-09-05
Sorbet syntax checker for ruby

tomm:


lcd047:
Please add a corresponding entry to the manual.

tomm:
I've added an entry in the manual :+1:

lcd047:
Thank you.  Merged with this patch on top: 3e31762a.

PR #2349, opened 2021-01-04
Fix govet errorformat

ohmystack:
Adapt for the latest govet

### DIFF
--- a/syntax_checkers/go/govet.vim
+++ b/syntax_checkers/go/govet.vim
@@ -23,7 +23,7 @@ function! SyntaxCheckers_go_govet_GetLocList() dict
     let makeprg = self.getExecEscaped() . ' vet'
 
     let errorformat =
-        \ '%Evet: %.%\+: %f:%l:%c: %m,' .
+        \ '%Evet: %f:%l:%c: %m,' .
         \ '%f:%l:%c: %m,' .
         \ '%W%f:%l: %m,' .
         \ '%-G%.%#'


ohmystack:
I'm using go1.13.12, the `go vet` error message looks like this:

```
vet: ./main.go:8:2: use of package fmt not in selector
```

I find that the old `errorformat` will parse this filename as `vet: ./main.go`, which should be `./main.go`, so I made this PR.
I'm not familiar with the regex format used by`errorformat`. If you have any ideas to optimize this PR, or have any docs about the format, please let me know. I'm glad to update my PR.

lcd047:
Ok, fixed in d97a664b.

ohmystack:
@lcd047  It would better if you can guide me, the PR's author, to fix this PR, instead of creating a new PR without mention me. To be honest, I'm not happy with this.

Syntastic is a famous vim plugin. I have used it for years, and I really like it. When this govet problem came to me, I spent much time to locate the problem and figure out how to fix it because I also use the vim-go and I have to tell which plugin caused the problem.

I appreciate your hard work of maintaining this great plugin. I want to share my advice about the workflow of treating PRs. I hope that the PRs in this project can work in a way like other famous open-source projects, embracing and respecting the contributors.

lcd047:
@ohmystack You want to be mentioned for a 1-line patch to a checker that exists only for legacy reasons (you're supposed to use [vim-go][vim-go] instead), in a Vim plugin that is essentially dead (it has been superseded by [ALE][ALE]).  Sure, please let me know how I should handle this to give you the recognition you deserve for your invaluable contribution.  Should I undo my patch so that you can post a new PR in your name?

PR #2364, opened 2021-05-19
Checker verilog/verilator: Update format.

xdch47:


lcd047:
Thank you.  Fixed in 176f364f.

PR #2372, opened 2021-07-19
Fix Erlang checker for test files with no test cfg

nickelization:
If we're editing a file in the "test" directory, the syntax checker
assumes we should use the 'test' profile and attempts to read from
"rebar.test.config" instead of "rebar.config". However, it is common to
have Erlang projects with tests that don't necessarily use a separate
"rebar.test.config" file, and in those types of projects the syntax
checker will break. This patch handles this case by falling back to
using "rebar.config" for tests if a separate "rebar.test.config" file is
not found.


lcd047:
Merged, thank you.

nickelization:
You're welcome! Thanks for the quick review and merge 😄

PR #2374, opened 2021-09-03
Checker go/golangci_lint: add cwd parameters and fix error format for some linters

xwind:
Some linters("typecheck", "lll") need pacakge path to work.
These linters were supported by golangci-lint, however can't work
correctly in Syntastic with lacking of package path.
You can check these supported linters list in https://golangci-lint.run/usage/configuration/

Morever, some linters ("lll", "goimports" etc) wouldn't pass column num to stdout. So, I add new errorformat pattern to recognize error messages from these linters.


xwind:
I didn't consider the compatibility with old args.

lcd047:
Merged, thank you.

PR #2375, opened 2021-09-06
Do not override options cppcheck arguments from config file with cppcheck_args

countsudoku:
Configuration for _cppcheck_ from a config file configured via `g:syntastic_cppcheck_config_file` was overwritten by content of `g:syntastic_cpp_cppcheck_args`. But for other checkers the options from the configured config file are appended to the call arguments of the checker. With this change the options from the _cppcheck_ config file will also be appended and not be overwritten by any static configuration.

But be aware, that this will change the behavior of the config options for _cppcheck_.

### DIFF
--- a/doc/syntastic-checkers.txt
+++ b/doc/syntastic-checkers.txt
@@ -693,8 +693,8 @@ Additionally:
                                           *'g:syntastic_cppcheck_config_file'*
 Type: string
 Default: unset
-File containing compilation flags (such as defines or include directories),
-one option per line (cf. |syntastic-config-files|).
+File containing additional compilation flags (such as defines or include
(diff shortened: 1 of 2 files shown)


lcd047:
If I recall correctly the idea at the time config files were added to `cppcheck` was to mimic what some Clang linters were doing -- they could use either a config file or command line options, not both at the same time.  The point was to have am easy way to turn off the config files, or some such.  But frankly that isn't particularly useful, so it wasn't implemented consistently to the other checkers.  So your patch makes complete sense.

On the other hand syntastic has been largely superseded by [ALE][ALE], it's only use these days is for people who are stuck with Vim 7.  It's probably way too late to make incompatible changes like this, they are likely to do more harm than good. _shrug_

PR #2376, opened 2021-09-10
Adds explicit buffer number to SyntasticStatuslineFlag.

SteveNeithardt:
When using splits in vim, it may happen that the statusline of
an inactive window gets redrawn. In that case, the 'current()'
buffer is no longer right. This fix finds the proper buffer
number to use, by using the g:statusline_winid when it exists.
In other cases, fallback to an empty current() call.

### DIFF
--- a/plugin/syntastic.vim
+++ b/plugin/syntastic.vim
@@ -695,7 +695,11 @@ endfunction " }}}2
 "
 "return '' if no errors are cached for the buffer
 function! SyntasticStatuslineFlag() abort " {{{2
-    return g:SyntasticLoclist.current().getStatuslineFlag()
+    if exists('g:statusline_winid')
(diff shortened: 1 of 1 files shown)


lcd047:
Nice detective work, but... no, sorry.  Syntastic has been superseded by [ALE][ALE], which takes advantage of Vim 8 features and is actively maintained.  The main use for syntastic these days is for people who are (for whatever reasons) stuck with Vim 7.  Sadly `statusline_winid` used in your patch has been introduced in Vim 8.1.1372.  Thus sorry, but no.

SteveNeithardt:
Oh ok. I'll admit to not checking version compatibility, my bad.

Good to know ALE exists. It could/should (?) be in the readme, so people that can upgrade to Vim 8 can also move on to ALE instead of staying here with Syntastic. I'll give it a go. Thank you for the review!

PR #2378, opened 2021-09-10
Add python3 checker

OmnipotentEntity:
What it says on the tin.  This adds an explicit "python3" syntax checker, that explicitly calls "python3" instead of "python," allowing an easy, system-wide workaround for distributions that are still shipping python2 as the default system python for legacy reasons.

### DIFF
--- a/syntax_checkers/python/codec3.py
+++ b/syntax_checkers/python/codec3.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python3
+
+from sys import argv, exit
+
+import codecs
+import re
+import os
+
+
+if len(argv) != 2:
+    exit(1)
+
+try:
+    with open(argv[1]) as fle:
+        text = fle.readlines()
+
+    if text:
+        match = re.match(r"#\s*coding\s*:\s*(?P<coding>\w+)", text[0])
+        if match:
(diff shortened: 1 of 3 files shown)


lcd047:
Please post patches only for problems that can't be solved without changing the code.  Also solutions that bypass the [existing mechanisms](https://github.com/vim-syntastic/syntastic/wiki/Syntax-Checker-Guide#duplicates) and duplicate code are not accepted.  Sorry, but no.

OmnipotentEntity:
Fair enough.  But I don't see how the existing mechanism you linked (which I understand to be "one checker for multiple filetypes") applies to this situation ("multiple checkers for one filetype").

Ultimately, I created this PR because I couldn't see a way to fix my issue without editing the code, but you are certainly more familiar with it than I am, so if you say there's a way I believe you.  However, I am still largely as confused as before.

lcd047:
> I don't see how the existing mechanism you linked (which I understand to be "one checker for multiple filetypes") applies to this situation ("multiple checkers for one filetype").

It doesn't: the solution here is to use virtual environments outside Vim, as described by the FAQ.  My point was that if you _ever_ find yourself duplicating checkers there's a better way around that.  And that still applies to duplicate checkers within the same filetype.

PR #2396, opened 2022-01-30
Add deprecation to README

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://github.com/vim-syntastic/syntastic/pull/2395


lcd047:
This is not a patch fixing a bug, is it. :smile: A deprecation note shall be added in due time.  That time is not now.

Sorry, but no.

